PR #2, opened 2019-06-28
dap and dap4 tests

lesserwhirls:
This PR does two things:

1. properly include dap4 support
2. update gretty configs such that we run dtswar and d4tswar from their published war files, which means we do not need that code in the netcdf-java repository so that we can test `httpservices`, `opendap`, and `dap4`.

Based on the branches `dap4` and `dap4.dmh`.


lesserwhirls:
Need to build on trusty for oraclejdk8 to work :-/

PR #14, opened 2019-07-05
Fix bug on Grib regular Time2D construction.

JohnLCaron:
Introduced during cleanup of 5.0, not present in 4.6.


JohnLCaron:
Not sure why .sh files got sent - must be line endings? Intellij wouldnt let me revert, probably user error.

Can resend if you need me too.

lesserwhirls:
I have that problem with .sh files sometimes, too, although I can never remember how I take care of them. I'll just commit them as you have them now and maybe we won't run into that issue in the future.

PR #16, opened 2019-07-06
Cdm-test works when netcdf4 c library is not installed.

JohnLCaron:
Also, more "problems" with git insisting that *.sh files are modified.


lesserwhirls:
I think I'm just going to remove those .sh files. They are in the git history if we need them, but they keep causing problems.

PR #17, opened 2019-07-07
Remove sleepycat (bdb) dependency.

JohnLCaron:
We have been using bdb for persistent storage, but only for GridDatasetInv.
Replace this with an in-memory cache.
Its possible we still want persistent disk storage, but its an optimization, so we need some data on what the need is to justify the complexity. We can add it back in later if needed. 

The TDS repo needs to be modified to remove references to MetadataManager. AFAICT, that repo is not yet on github. Ok to wait on this PR for that if you want.

Also ok to investigate impact before removal if you want.


lesserwhirls:
Let me chew on this one for a bit. I'm working on UGRID support in (ultimately) FMRCs and want to have that ready for the 5.0 release of the TDS. I'm a little worried that any issues that arise from this change will prevent IOOS nodes from upgrading to 5 for the UGRID support. I'm trying to get a real-time data feed of UGRID stuff so that we can have some local FMRCs to better kick the tires of these kinds of changes, but so far I've not had any luck obtaining a feed.

However, longer term (year time frame) I will be working to create a GRID feature collection (using the new coverage stuff), that keeps the functionality of FMRCs when working with coverage datasets that have run times and valid times. Once that's working, it might make sense to deprecate FMRCs?

JohnLCaron:
To clarify, this PR doesnt change any functionality, it removes the persistant cache, which would effect only performance. It keeps an in-memory cache, which gets discarded when the program exits.

My thought is to measure performance and then add a persistent cache back in, if needed, perhaps only on the server? Maybe use a pluggable persistance layer?

JohnLCaron:
Im thinking that instead of putting these into a key-value store, just write GridDatasetInv.xml next to the original, a la grib collections. 

It would be good to rework FMRC along the lines of GRIB collections, and possibly design a new summary format and GRID object model, if youre up for it.

Then on the server, use an object cache for speed.

lesserwhirls:
> My thought is to measure performance and then add a persistent cache back in, if needed, perhaps only on the server? Maybe use a pluggable persistance layer?

I think that is what is giving me pause (server side performance), at least for the inital 5.0 release. 

> Im thinking that instead of putting these into a key-value store, just write GridDatasetInv.xml next to the original, a la grib collections.

One thing I've been pondering is the ability for the TDS to move to a microservice and/or serverless architecture. In that case, a key-value store (controlled by a pluggable persistence layer) might be a better way to go?

> It would be good to rework FMRC along the lines of GRIB collections, and possibly design a new summary format and GRID object model, if youre up for it.

I think there are bits and pieces in the current `.ncx4`  (axis types, collection partitions, etc.), as well as some of the GRIB collection machinery, that could be abstracted out and used for this. Of course, that's quite possibly a very naive and dangerous view on my part.

lesserwhirls:
Ok, let's go ahead and do this. Not sure how bad ripping it out of the TDS will be, but let's not let that hold us up on netCDF-Java 5.

PR #30, opened 2019-07-24
A temporary change troubleshooting a Jenkins test hang, seeing an infinite loop.

hvandam2:
A temporary change troubleshooting a Jenkins test hang, seeing an infinite loop
after "InterruptedException=sleep interrupted" during TestNetcdfFileCache > testPeriodicClear.
Adding some output statements to dump stacktraces to the console.


lesserwhirls:
Ok, I think we're good for a rebase on this branch.

PR #37, opened 2019-07-27
fix "critical bugs" from sonarcloud.

JohnLCaron:
Fix 7/8 bugs. The other one is false positive.


lesserwhirls:
@dopplershift - would you mind taking a look at `cdm/src/main/java/ucar/nc2/iosp/nexrad2/Level2VolumeScan.java`? There are a lot of formatting changes, but a few others as well.

dopplershift:
@lesserwhirls I think those changes look reasonable.

lesserwhirls:
Thanks @JohnLCaron @dopplershift!

PR #45, opened 2019-08-05
Fix main to support whitespaces at filename

cofinoa:
The `NCdumpW.main` doesn't support filenames with escaped or quoted whitespaces. 
```shell
$ java ucar.nc2.NCdumpW "File Name With Whitespaces.nc" 
file not found= File
```
This PR fix this issue. 

Probably this fix needs to be extended to the rest of arguments.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/NCdumpW.java
+++ b/cdm/src/main/java/ucar/nc2/NCdumpW.java
@@ -693,17 +693,25 @@ public class NCdumpW {
       System.out.println(usage);
       return;
     }
-
-    StringBuilder sbuff = new StringBuilder();
-    for (String arg : args) {
-      sbuff.append(arg);
-      sbuff.append(" ");
-    }
-
     try {
       Writer writer = new BufferedWriter(new OutputStreamWriter(System.out, CDM.utf8Charset));
-      NCdumpW.print(sbuff.toString(), writer, null);
-
+      // pull out the filename from the command
+      String filename = args[0];
+      ucar.nc2.util.CancelTask ct=null;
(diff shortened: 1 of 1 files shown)


cdm/src/main/java/ucar/nc2/NCdumpW.java
```diff
@@ -693,17 +693,25 @@ public static void main(String[] args) {
...
+      ucar.nc2.util.CancelTask ct=null;
+      try (NetcdfFile nc = NetcdfDataset.openFile(filename, ct)) {
+        // the rest of the command
+        StringBuilder command = new StringBuilder();
+        for (int i=1; i < args.length; i++) {
+          command.append(args[i]);
+          command.append(" ");
+        }
+        print(nc, command.toString(), writer, ct);
+    } catch (java.io.FileNotFoundException e) {ls -la
```
lesserwhirls: `ls -la` and `:wq` end up in my code pretty regularly.

lesserwhirls:
There are other command line utilities that likely have the same issue. At least one of them, `nccopy`, uses `jcommander`, and in the process, I *think* it handles paths with spaces:

https://github.com/Unidata/netcdf-java/blob/master/cdm/src/main/java/ucar/nc2/write/Nccopy.java

It'd be nice to get all of them using `jcommander`, if that fixes this issue...but for now, I'll add it to the to-do list :-)

cofinoa:
Probably there are more tools, but Nccopy doesn't suffer from this issue.

JCommander appears to be good solution, but docopt implementation for java it's more declarative:

https://github.com/docopt/docopt.java

lesserwhirls:
Fixed up and rebased in Unidata/netcdf-java#134. Will be in the next release. Thank you @cofinoa!

PR #47, opened 2019-08-07
Intellij inspections

JohnLCaron:
These are (mostly) automatic Intellij fixes.
Do not merge until we have tested against TDS.
Problem is that we have removed some unneeded checked exceptions, so this could break compilations.


JohnLCaron:
Maybe this should go into 5.1?

JohnLCaron:
done

lesserwhirls:
Having given it some more though, I think we should bump the version to 5.1. I can do that in another PR though, so I'll go ahead and merge this in.

JohnLCaron:
what exactly sets the version?

lesserwhirls:
The version is set in build.gradle:

https://github.com/Unidata/netcdf-java/blob/136a8d3dc0978c82067e56a3a1117516cc33b58d/build.gradle#L41

JohnLCaron:
have we tested against tds?

lesserwhirls:
This PR passed the travis tests for tds. Once the version of netCDF-Java is bumped in the TDS dependency file (PR waiting to pass travis), I'll run the full set of tests on Jenkins.

PR #48, opened 2019-08-08
Wrong Immutable class (no longer in Java 11).

JohnLCaron:


JohnLCaron:
Compiling with Java 11 uncovered this problem. Wont affect Java 8 users.

PR #49, opened 2019-08-08
Test TDS against netcdf-java PRs on travis-ci

lesserwhirls:


lesserwhirls:
The fact that this addition seemed to "work" on the first try makes me nervous, but hey, a broken clock is right two times each day, right?

nervous

dopplershift:
Honestly, I'd try putting in a PR that you know would break it and see what happens...

PR #55, opened 2019-08-09
remove unneeded throws Exceptions

JohnLCaron:
This is part 2 of intellij cleanup.
Finish removing unneeded checked Exceptions, except leave them on top-level interfaces (even restore a few that had been removed).  This should also cut down on the affects on library users.

Probably more to think about when checked Exceptions should be used.


lesserwhirls:
Now we know the TDS travis check works 👍

lesserwhirls:
I'm happy to merge this in, knowing that there will be some TDS fixes to make to get it working again. But, at least we know ahead of time!

PR #57, opened 2019-08-14
Fix cdm tests.

JohnLCaron:
Some miscellaneous cdm-test fixes.
Tried to work with Categories; may not be working fully with IntelliJ-launched unit tests (?)
Also some changes to files in /cdmUnitTest/ eg 
/cdmUnitTest/conventions/cfradial/cfrad.20171127_202111.203_to_20171127_202123.085_DOW7_v275_s04_el7.00_SUR.nc
Can I get at Unidata's copy o those files, or do you have to send to me??

### DIFF
--- a/gradle/root/testing.gradle
+++ b/gradle/root/testing.gradle
@@ -42,6 +42,7 @@ ext {
     }
     
     isRdaDataAvailable = false  // figure out how to set later
+    runSlowTests = false  // figure out how to set later
 }
 
 import java.nio.file.*
--- a/cdm-test/src/test/java/ucar/nc2/dataset/TestCFRadial.java
+++ b/cdm-test/src/test/java/ucar/nc2/dataset/TestCFRadial.java
@@ -50,6 +50,7 @@ public class TestCFRadial {
     private RadialDatasetSweep arrayLatLonData() throws IOException
     {
         String filename = TestDir.cdmUnitTestDir + "conventions/cfradial/cfrad.20171127_202111.203_to_20171127_202123.085_DOW7_v275_s04_el7.00_SUR.nc";
+        System.out.printf("arrayLatLonData= %s%n", filename);
         Formatter buf = new Formatter();
         return (RadialDatasetSweep) FeatureDatasetFactoryManager.open(
                 FeatureType.RADIAL, filename, null, buf);
--- a/cdm-test/src/test/java/ucar/nc2/grib/TestCoordinatesMatchGbx.java
+++ b/cdm-test/src/test/java/ucar/nc2/grib/TestCoordinatesMatchGbx.java
@@ -19,11 +19,13 @@ import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
+import ucar.unidata.util.test.category.Slow;
 
 /**
  * Test reading grib coordinates match gbx
  */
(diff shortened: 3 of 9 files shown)


JohnLCaron:
Does Jenkins run cdm-test? Can I find out how it did?

lesserwhirls:
Hi @JohnLCaron - sorry for the delay. Here is a copy of that file (zip compressed):

[cfrad.20171127_202111.203_to_20171127_202123.085_DOW7_v275_s04_el7.00_SUR.nc.zip](https://github.com/Unidata/netcdf-java/files/3503457/cfrad.20171127_202111.203_to_20171127_202123.085_DOW7_v275_s04_el7.00_SUR.nc.zip)

We tried hosing the test datasets on git lfs, but it more than doubled the size of the test datasets on disk due to the way git works, so it was a no-go. Hopefully we'll have the test data publicly accessible in the next week or so. I'm happy to add/remote/update/fetch any datasets you need until we get that up and running.

Jenkins runs the full test suite, including the stuff in cdm-test. I've tried running your PR twice today, and both times we are getting hung up on `ucar.nc2.util.cache.TestNetcdfFileCache.testPeriodicClear()`:

~~~bash
ucar.nc2.util.cache.TestNetcdfFileCache > testPeriodicClear STANDARD_OUT
     InterruptedException=sleep interrupted
     InterruptedException=sleep interrupted
     InterruptedException=sleep interrupted
~~~

We've had this problem periodically, and recently we added some debug statements to try to diagnose it...however, none of those messages showed up, so we didn't hit up the correct places. It does not happen consistently, unfortunately (never been able to catch it in the debugger). I'll give it a spin again and we'll see.

BTW - the Jenkins github authentication plugin has been fixed, so soon you'll be able to use that to access Jenkins. At that point, you'll be able to run tests on your branches whenever you need to.

gradle/root/testing.gradle
```diff
@@ -42,6 +42,7 @@ ext {
     }
     
     isRdaDataAvailable = false  // figure out how to set later
+    runSlowTests = false  // figure out how to set later
```
lesserwhirls: So for the Travis or Jenkins category, we inspect the systems environmental variable to see if we are running in those environments as part of the gradle build. If so, then we set the gradle variables `isJenkins` and `isTravis` appropriately and let gradle exclude the appropriate categories. For example, if we're on Jenkins, skip anything in the category `ucar.unidata.util.test.category.NotJenkins`, and if we're on Travis skip anything in the categories `ucar.unidata.util.test.category.NotTravis`, or `ucar.unidata.util.test.category.NeedsExternalResource`. For this, we need to decide when we want to run the slow tests.

In this case, it might be best to allow runSlowTests to be set with a property we pass through gradle, say `./gradlew -DrunSlowTests=false testAll`, but we'll need to add some code to `gradle/root/testing.gradle` to look for that and set the variable properly.

Now, if you're running the test through intellij (directly through the junit run/debug option), the Category machinery in the gradle build will be ignored and the tests will just run, unless you explicitly tell junit to exclude categories by passing in `-DexcludedGroups=ucar.unidata.util.test.category.Slow` as a vm option...I think...

lesserwhirls:
I'm going to go ahead and merge this PR, and will add an issue about enabling the slow category. I can take care of that.

JohnLCaron:
its taking >30 minutes on my machine. could also leave it off by default.
up to you.

PR #59, opened 2019-08-14
add wrf to gradle

piqueen314:
add wrf module to gradle 
also kind of a test pull reruest


dopplershift:
So you want:
```
git commit --amend
```
and
```
git push --force
```

PR #62, opened 2019-08-16
fix "high impact" Coverity defects

JohnLCaron:


JohnLCaron:
So theres a failure in tds, due to an API change in netcdf-java.
Not sure how to do a PR for that, since i need the new version of netcdf-java. 
Lets figure out a good workflow for this, since it will happen a lot.

Seems like we want to do a SNAPSHOT release of netcdf-java, which I then use in tds??

JohnLCaron:
I see that tds depends on 
  versions["ncj"] = "5.1.0-SNAPSHOT"
so maybe this is already set to work, and just need to push the snapshot??
or maybe we need to use a minor version 5.1.x each time this happens ??

lesserwhirls:
So there are a few ways to handle the failure in the TDS build.

Option 1: Preemptively get a PR ready to roll for the TDS

To do this, you'll need to do a few things. First, publish a local snapshot of the netcdf-java artifacts to your local maven repository (.m2 directory under your users home directory). You can do that using:

~~~java
./gradlew clean publishToMavenLocal
~~~

Next, make a branch in your tds repository to hold the fixes. First thing to do there is to edit
* build.gradle
* gradle/any/dependencies.gradle

and uncomment the `//mavenLocal()` lines (one in each). This will tell the tds build to look in your local maven repository for artifacts first, and will pick up your local snapshot. Then you're on your way to fixing the TDS issues. Once that is all done, I can merge this PR, trigger a Jenkins run (which will deploy the new snapshot to our nexus server). After that's done, you can recomment the `mavenLocal()` lines in the TDS gradle files, and submit the PR to TDS.

Option 2: Let's merge this is, let Jenkins deploy a new snapshot (that will break TDS builds temporarily), then work on getting the TDS errors fixed (I can take a stab at that).

To me, option 2 seems easiest, but has the risk of breaking other TDS developer environments. On one hand, the TDS is using a snapshot of netCDF-Java, and if we were using a snapshot for any other library, we'd be asking for it, so to say. With the new travis test, at least we have a heads-up that there will be an issue.

Option 3: Explicitly use a dated snapshot version of the the netCDF-java artifacts in the TDS build, and add one more `sed` command to the netCDF-Java Travis build process (for TDS tests) to replace that with the generic snapshot, thus picking up and using the netCDF-Java artifacts generated by the netCDF-Java PR. Then we can use option 2 without the risk of breaking other dev environments, and as part of the PR against the TDS repo for fixing the identified issues in the netCDF-Java travis run, we can bump to a newly dated snaphot.

I'd say option 3.

lesserwhirls:
But for this PR? Option 2 will work for now.

JohnLCaron:
ok

JohnLCaron:
Only the tds unit tests are failing at this point.

lesserwhirls:
Merging, and I'll fix TDS today. Thanks @JohnLCaron!

lesserwhirls:
Can't do option 3 until we upgrade to the latest version of gradle. Before gradle 5.6, artifacts were published using the timestamp of when the artifact was assembled. This means that each a specific artifact has a different version string. For example, these artifacts were generated from the same repository:

~~~
bufr-5.1.0-20190819.181329-7.jar
cdm-5.1.0-20190819.181333-7.jar
grib-5.1.0-20190819.181353-7.jar
~~~

In order to use this specific version (20190819 build 7) in the TDS gradle build, we'd need to keep track of every artifacts specific version string. Yuck.

As of 5.6 [see here](https://github.com/gradle/gradle/issues/1772), the same timestamp is used in the version name of a SNAPSHOT across all submodules, which is what we'd want to have.

For now, I'm happy to do the leg work of option 1 or 2 on the TDS side as things break.

PR #63, opened 2019-08-17
upgrade packages including protobuf and jna. Regenerate proto code.

JohnLCaron:
Notable issues: 
1. protobuf upgrade to 3.9.1
PointStreamProto.Member conflicts with NcStreamProto.Member, so I changed it to PointStreamProto.PointFeatureMember. If this is used by python library, will have to coordinate.

2. jna upgraded to 5.4.0
There were a few changes needed in Nc4*.java. Might want to let Dennis know, though I doubt theres any problem.

### DIFF
--- a/bufr/src/main/java/ucar/nc2/ft/point/bufr/BufrCdmIndexProto.java
+++ b/bufr/src/main/java/ucar/nc2/ft/point/bufr/BufrCdmIndexProto.java
@@ -6,7 +6,13 @@ package ucar.nc2.ft.point.bufr;
 public final class BufrCdmIndexProto {
   private BufrCdmIndexProto() {}
   public static void registerAllExtensions(
+      com.google.protobuf.ExtensionRegistryLite registry) {
+  }
+
+  public static void registerAllExtensions(
       com.google.protobuf.ExtensionRegistry registry) {
+    registerAllExtensions(
+        (com.google.protobuf.ExtensionRegistryLite) registry);
   }
   /**
    * Protobuf enum {@code FldAction}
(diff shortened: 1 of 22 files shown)


JohnLCaron:
I see this comment in tds dependencies.gradle:

libraries["chronicle-map"] = dependencies.create("net.openhft:chronicle-map:3.17.0") {
    // the version of jna shipped with chronicle-map requires GLIBC 2.14
    // which prevents the use of the TDS on several versions of linux, such
    // as CentOS and Hed Hat 6. We already include jna in netCDF-Java for
    // netCDF-C support, so we will rely on that version instead.
    // We should be able to get rid of this on the next release of chronicle-map,
    // as their snapshot BOM pom shows that they pulled back on jna.
    exclude group: 'net.java.dev.jna', module: 'jna'
    exclude group: 'net.java.dev.jna', module: 'jna-platform'
    // We do not need XStream for our use of chronicle-map.
    // Nuke it. [SEC]
    exclude group: 'com.thoughtworks.xstream', module: 'xstream'
    // We do not need pax-url-aether
    // Nuke it. [SEC]
    exclude group: 'org.ops4j.pax.url', module: 'pax-url-aether'
}

Its likely that this is the version of jna this PR upgrades to.
Theres lots of chatter about this everywhere. Oh yeah, its been going on for at least 6 years. 
Maybe its time to just run on the latest versions?

lesserwhirls:
RHEL and CentOS 6 will stop receiving updates at the end of November 2020. When this came up last year, I was a bit reluctant to take that stance For 5.0. However, at this point, I think we can and should go forward.

HOWEVER...that might not be an issue now. It looks like JNA 5.4.0 uses a minimum GLIBC of 2.7, so it might not even be a problem anymore?

In terms of python and `PointStreamProto.Member`...@dopplershift?

JohnLCaron:
Ok, sounds like not a problem....

PR #77, opened 2019-08-23
Spotless

lesserwhirls:


lesserwhirls:
Ok, so the eclipse style file has `join_lines_in_comments=true`, which will join lines in comments up to the column width:

https://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fjdt%2Fcore%2Fformatter%2FDefaultCodeFormatterConstants.html


So we see things like this. Before:

~~~java
/**
 * Manage cdm index (ncx) for Bufr files.
 * Covers BufrCdmIndexProto
 * Never completes or in operational use, could redo as needed
~~~

after:

~~~
/**
 * Manage cdm index (ncx) for Bufr files. Covers BufrCdmIndexProto Never completes or in operational use, could redo as
 * needed
~~~

I am going to make this false.

lesserwhirls:
Ok, I think that has it.

PR #79, opened 2019-08-25
Fix test failures in cdm-test

JohnLCaron:


JohnLCaron:
thanks for cleaning up....

PR #82, opened 2019-08-31
Clean up cruft associated with System.out.println

JohnLCaron:


lesserwhirls:
Hey @JohnLCaron - sorry I didn't ask this before the merge, but what was the thought behind moving the cinrad IOSP classes from `cdm` to `legacy`?

JohnLCaron:
Sorry, it was probably a mistake to do that.

I was in some kind of fever to clean things up, and that one has so much
cruft, plus its documentation is completely wrong, it has no references to
what it actually is, no docs etc.
I was imagining its IDV specific and maybe something we can leave behind in
version 6.

However, it got mixed up with the more mechanical cleanups, and I should
have left it off until I could check with you, and even then it should be
in its own PR.

So Im good with moving that back until we can consider it more, or whatever
you want to do.

lesserwhirls:
I'll move it back for now, but I think it would be a good target for the yet-to-be-created `radar` module :-) The reason I noticed is that we have a TDS integration test failing because of it on Jenkins. Not sure that we have any TDSs serving out cinrad data or not, but maybe.

PR #83, opened 2019-08-31
Missorted modifiers

JohnLCaron:
Apply this inspection across all files in netcdf-java.


JohnLCaron:
Not sure how a spotless mis-format got introduced here. If you have any insight, let me know.
Also, I accidentally committed on the wrong branch, so if its easy for you to run spotlessAppply on this branch, please do.

Btw, can we apply spotlessAppl automatically, if we are going to require it anyway?

lesserwhirls:
Rebased and applied spotless.

> Btw, can we apply spotlessAppl automatically, if we are going to require it anyway?

We could implement a git pre commit hook so that spotlessApply runs when you initial a commit with `git commit`, but before the commit is actually made. I'll take a look at that.

PR #95, opened 2019-09-07
Remove unused dependencies.

JohnLCaron:
Remove imgscalar dependency by bringing in the source.

PR #96, opened 2019-09-07
Some cleanup on dt.image

JohnLCaron:


lesserwhirls:
This should take care of Unidata/netcdf-java#95 as well.

PR #100, opened 2019-09-14
remove server-only code from opendap client library.

JohnLCaron:
For 5.2: do not merge until 5.1 is out.


JohnLCaron:
Just failing on TDS I think. 
There a PR on TDS that should fix that.

lesserwhirls:
Good to go here. 5.1 is out (shipped it last Thursday), but I have not made the announcement yet.

PR #102, opened 2019-09-14
Cdm sub modules

JohnLCaron:
pass 1, do not commit yet.


JohnLCaron:
What should we do to get the train unstuck?

lesserwhirls:
So we're passing the netcdf-java tests, but there is a conflict - can you rebase and run spotless? I'm ok merging with the TDS failing. I have it compiling, but the changes that Dennis made are not playing well with the TDS tests. I have an idea about what is going on, but it will take a little more time to tease out.

lesserwhirls:
Fixed up on netcdf-java#124 - if that one looks good to you, then I'll close this and merge that one in.

JohnLCaron:
Do you want me to fix up TDS dependencies to use 5.2?
Or does TDS 5.1 need to be released still?

lesserwhirls:
TDS is still at 5.0 as a beta, and is currently using `5.2.0-SNAPSHOT` for the netcdf-java dependencies. Once these changes go in, I'll publish a new 5.2.0 snapshot with the changes, and then I can take a pass at fixing up the TDS dependencies. Our TDS test will still fail due to changes made in `httpservices`, but I'm about to open an issue for Dennis to take a look. Credential caching changed and I'm having trouble figuring out out to modify the test utilities to work with the new caching. I might just add the nottravis annotation to those TDS tests until it gets fixed.

JohnLCaron:
So im making changes that have some effect on TDS. I think the right thing
to do is to bring TDS along as NJ evolves, and not accept PR until TDS
builds, I presume that would have caught the problem with credential
caching. So im in favor of fixing temporarily the breakage, then making
sure TDS doesnt break at each PR.

I would think you would want a TDS 5.1 release that uses NJ 5.1. I have a
PR im working on (hide apache http dependence), and it makes a lot of sense
to let me fix the small breakages from it. But then TDS needs to be changed
to use 5.2 dependencies. etc.

Let me know how you want to proceed...

lesserwhirls:
> So im making changes that have some effect on TDS. I think the right thing to do is to bring TDS along as NJ evolves, and not accept PR until TDS builds, I presume that would have caught the problem with credential caching. So im in favor of fixing temporarily the breakage, then making sure TDS doesnt break at each PR.

I just made and merged a PR against the TDS so that it will build with the current master of netcdf-java (5.2-SNAPSHOT). There were compile issues, but those are addressed now. There were also new test failures, but it's due to the way the credential caching is being handled now in our test suite. I cannot quite get our test specific code to do the right thing (looks like we had some special test only code to deal with caching, which might not be needed anymore). The failure wasn't actually anything related to the TDS itself (admin/debug still works through the browser, etc.). However, this does point to a hole in the httpservice testing on the netcdf-java side that was introduced during the split. 

>  I would think you would want a TDS 5.1 release that uses NJ 5.1.

We're still on 5.0.0-SNAPSHOT for TDS (we've had beta releases, but not stable yet). I don't think the versions of netcdf-java and tds need to be in lockstep now that they are split.

> I have a PR im working on (hide apache http dependence), and it makes a lot of sense to let me fix the small breakages from it. But then TDS needs to be changed to use 5.2 dependencies. etc. Let me know how you want to proceed...

The TDS has been using 5.2.0-SNAPSHOT of the netcdf-java dependencies since we bumped. I've been bumping the snapshots used by the tds as we go until I get to the point of making the TDS 5 release. I'm thinking that we'll be at the 5.2 release of netCDF-Java before TDS 5.0 release ships.

I have made changes to the TDS to use the new modules, but I have a question about the naming of the new module that'd I'd like to take care of before I make the PR (see https://github.com/Unidata/netcdf-java/pull/123#issuecomment-533818016)

PR #117, opened 2019-09-17
Fix UrlAuthenticatorDialog.java to avoid repeated dialog popups.

DennisHeimbigner:
TdsMonitor was repeatedly prompting for credentials for each
type of log it downloads (and every single thredds servlet log
downloaded). Additionally, it was locking up if a bad password
was provided.

The repeated popup problem was occuring because
UrlAuthenticatorDialog had been modified at one time to cache
credentials that it generated so that repeated calls would
return the cached credentials.  This caching is required by
Apache httpclient as of at least 4.5.x.  It is used, for
example, to store proxy credentials. The canonical code example
is to look at the code for BasicCredendentialsProvider.  In any
case, someone removed this code at some point. No idea why.

With respect to the problem of a bad username+password, I spent some
time looking into the Apache code. It turns out that the
AuthCache object in the HttpContext object is notified when authorization
fails. The AuthCache.remove method is called with the HttpHost object
defining the host for which authorization failed. It turns out that
AuthCache does not, by default, notify the CredentialsProvider object
of the failure (a serious flaw IMO).

So, I have added a new class -- HTTPAuthCache -- that can
store a set of CredentialsProvider instances and when the remove()
method is called, it forwards the call to each CredentialsProvider instance
using one of two methods. If the CredentialsProvider implements
a new Interface -- HTTPCredentialsProvider -- then the remove()
method of the HTTPCredentialsProvider API is called. Otherwise,
the clear() method of the CredentialsProvider is called; this latter
case is provide backward compatibility. The former case provides
the HTTPCredentialProvider with the ability to carry out a finer
grain removal of bad credentials.

This PR does the following:
- Simplified the HTTPSession credentials interface to just
  expose setGlobalCredentialsProvider and setCredentialsProvider
  (the per-instance case). All of the setCredentials() methods
  are removed from HTTPSession and it is assumed that the the
  user will use the setCredentials method of CredentialsProvider
  instead. This assumes that the CredentialsProvider object
  supports credentials caching.
- Added HTTPAuthCache.java and HTTPCredentialsProvider.java.
- Modified HTTPSession to use HTTPAuthCache instead of BasicAuthCache
  in order to forward remove() method calls to the global and per-instance
  CredentialsProviders.
- Restore the caching behavior to UrlAuthenticatorDialog
  and make it implement the HTTPCredentialsProvider interface
  instead of the CredentialsProvider interface. This allows
  for finer grain removal of bad credentials.
- Removed the HTTPProviderFactory as no longer being necessary and because
  it added unneeded complexity.
- Modified TestHTTPSession to conform to the above changed.
- Fixed a bug in HTTPAuthUtil with respect to conversions between
  an AuthScope instance and an HttpHost instance. The scheme handling
  was incorrect.
- Updated the documentation.

Notes:

1. It is possible that if the server redirects authorization handling
   to a different server, then the host argument to remove might
   be the redirected server rather than the original server. In this
   case, credentials removal might fail. This has not been tested.
2. This PR has not been tested by either Jenkins (no permissions)
   nor Travis.


lesserwhirls:
Thank you @DennisHeimbigner! You PR runs on Travis once you make the PR. The netcdf-java tests passed, but the tds test failed. That's ok, as it is intended to give us a heads-up that there will be issues. I'll go ahead and merge this in and I can take a look at the tds failures tomorrow. That will give me a chance to try out using the changes you've made here.

PR #118, opened 2019-09-18
protocol service provider strawman

JohnLCaron:
First pass "strawman" on using ServiceLoader for optional protocols, to get your input.
Do HDFS and S3 fit into this?

### DIFF
--- a/cdm/src/main/java/thredds/client/catalog/tools/CatalogNetcdfFileProvider.java
+++ b/cdm/src/main/java/thredds/client/catalog/tools/CatalogNetcdfFileProvider.java
@@ -1,0 +1,27 @@
+package thredds.client.catalog.tools;
+
+import java.io.IOException;
+import java.util.Formatter;
+import thredds.client.catalog.ServiceType;
+import ucar.nc2.NetcdfFile;
+import ucar.nc2.dataset.DatasetUrl;
+import ucar.nc2.dataset.NetcdfFileProvider;
+import ucar.nc2.util.CancelTask;
+
+public class CatalogNetcdfFileProvider implements NetcdfFileProvider {
+
+  @Override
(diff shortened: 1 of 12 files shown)


JohnLCaron:
Could also refactor ncml, cdmremote and thredds out of NetcdfDataset with this mechanism.

lesserwhirls:
So HDFS and S3 work at the `RandomAccessFile` level, &agrave;l la `HTTPRandomAccessFile` and get hooked-up inside `NetcdfFile` (via `getRaf`). For those (and other extensions of `RandomAccessFile`), perhaps a similar approach could be taken (say, `RandomAccessFileProvider`)?

Then again, we can control what `RandomAccessFile` class is actually used at the `NetcdfFileProvider` level with `open`. At that point, we could pull all of the `RandomAccessFile` extensions out of NetcdfFile and refactor `NetcdfFile.getRaf()` to use the new `NetcdfFileProvider` infrastructure that you have here (would probably want to bump `netcdfFileProvider` up a level into `ucar.nc2`).

Either way, I like the way this is looking!

JohnLCaron:
There has to be some way of telling the library to use the special RandomAccessFile, so I think that means it has to be a protocol service provider. Presumably they can use 

protected NetcdfFile(IOServiceProvider spi, ucar.unidata.io.RandomAccessFile raf, String location,
      ucar.nc2.util.CancelTask cancelTask)

So at the moment I think this is right way to handle HDFS and S3.

JohnLCaron:
Since the goal is to allow these to be added without touching core code, I see that DatasetUrl needs to be modified to look for these providers, so that their protocol string doesnt have to be hardcoded there. So using enum ServiceType has to be generalized. So a bit more work to be done here.

JohnLCaron:
Superceded by #128

PR #127, opened 2019-09-22
Hide (mostly) apache.http other than httpservices.

JohnLCaron:

PR #121, opened 2019-09-20
Allow protocol service providers that do not have a ServiceType.

JohnLCaron:

PR #128, opened 2019-09-22
Protocol service providers

JohnLCaron:
This follows #127 and supercedes #118 and #121, which will be deleted after this one has made it through.


JohnLCaron:
Not sure why TDS TestReify  is failing since it was fixed in tds PR45.
Somehow this PR is using an older version??

JohnLCaron:
included in #129

PR #138, opened 2019-10-21
Adding a System Property for the maximum HTTP Buffer size.

hvandam2:
The property name is: ucar.unidata.io.http.maxHTTPBufferSize and will default to 10000000.


lesserwhirls:
Thanks @hvandam2!

Issue #140, opened 2019-10-25
Reading web netCDF data: Connection timed out when looking for https resource (but it timed out looking for http resource)


lesserwhirls:
From the [mailing list](https://www.unidata.ucar.edu/mailing_lists/archives/netcdf-java/2019/msg00022.html)

---

When using `cdm-core-5.2.0`, when attempting to access a netCDF resource on a server that serves https on port 443 but does not serve http on port 80, attempting to access the https resource fails but with a message indicating it tried port 80.

See redacted stack trace below.

I wonder if it is related to the diff at https://github.com/Unidata/netcdf-java/compare/v5.1.0...v5.2.0#diff-89b73930910f1f42a6af87d6d282a372 but this is speculation.

~~~
Failed to open netCDF file https://fqdn-of-server/path/to/netcdf_resource.nc
...
Caused by: ucar.httpservices.HTTPException: org.apache.http.conn.HttpHostConnectException: Connect to fqdn-of-server:80 [fqdn-of-server/x.x.x.x] failed: Connection timed out: connect
        at ucar.httpservices.HTTPMethod.executeRaw(HTTPMethod.java:373)
        at ucar.httpservices.HTTPMethod.execute(HTTPMethod.java:314)
        at ucar.unidata.io.http.HTTPRandomAccessFile.doConnect(HTTPRandomAccessFile.java:136)
        at ucar.unidata.io.http.HTTPRandomAccessFile.<init>(HTTPRandomAccessFile.java:60)
        at ucar.unidata.io.http.HTTPRandomAccessFile.<init>(HTTPRandomAccessFile.java:40)
        at ucar.nc2.NetcdfFile.getRaf(NetcdfFile.java:448)
        at ucar.nc2.NetcdfFile.open(NetcdfFile.java:338)
        at ucar.nc2.NetcdfFile.open(NetcdfFile.java:305)
        at ucar.nc2.NetcdfFile.open(NetcdfFile.java:290)
        at ucar.nc2.NetcdfFile.open(NetcdfFile.java:278)
        at wres.io.reading.nwm.NWMTimeSeries.openFile(NWMTimeSeries.java:194)
        ... 52 more
Caused by: org.apache.http.conn.HttpHostConnectException: Connect to fqdn-of-server:80 [fqdn-of-server/x.x.x.x] failed: Connection timed out: connect
[...]

Issue #120, opened 2019-09-18
cdmremote fails on https: URL


JohnLCaron:
works:
http://thredds.ucar.edu/thredds/cdmremote/grib/NCEP/GFS/Pacific_20km/GFS_Pacific_20km_20190918_1200.grib2

fails:
https://thredds.ucar.edu/thredds/cdmremote/grib/NCEP/GFS/Pacific_20km/GFS_Pacific_20km_20190918_1200.grib2

error message:
ucar.httpservices.HTTPException: java.net.URISyntaxException: Expected scheme-specific part at index 5: HTTP:
	at ucar.httpservices.HTTPAuthUtil.authscopeToURI(HTTPAuthUtil.java:112)
	at ucar.httpservices.HTTPSession.init(HTTPSession.java:811)
	at ucar.httpservices.HTTPSession.<init>(HTTPSession.java:797)
	at ucar.httpservices.HTTPFactory.newSession(HTTPFactory.java:38)
	at ucar.nc2.stream.CdmRemote.<init>(CdmRemote.java:79)
	at ucar.nc2.stream.CdmRemoteNetcdfFileProvider.open(CdmRemoteNetcdfFileProvider.java:19)
	at ucar.nc2.dataset.NetcdfDataset.openOrAcquireFile(NetcdfDataset.java:712)
	at ucar.nc2.dataset.NetcdfDataset.openDataset(NetcdfDataset.java:430)
	at ucar.nc2.dataset.NetcdfDataset.acquireDataset(NetcdfDataset.java:576)
	at ucar.nc2.dataset.NetcdfDataset.acquireDataset(NetcdfDataset.java:536)
	at ucar.nc2.ui.ToolsUI.openFile(ToolsUI.java:1271)
	at ucar.nc2.ui.op.DatasetViewerPanel.process(DatasetViewerPanel.java:98)
	at ucar.nc2.ui.OpPanel.doit(OpPanel.java:173)
	at ucar.nc2.ui.OpPanel.lambda$new$0(OpPanel.java:83)
	at javax.swing.JComboBox.fireActionEvent(JComboBox.java:1258)
	at ucar.ui.prefs.ComboBox.fireActionEvent(ComboBox.java:160)
	at javax.swing.JComboBox.setSelectedItem(JComboBox.java:586)
	at javax.swing.plaf.basic.BasicComboBoxUI$Handler.actionPerformed(BasicComboBoxUI.java:1943)
[...]

PR #141, opened 2019-10-26
Fix use of AuthScope.getScheme()

DennisHeimbigner:
re: https://github.com/Unidata/netcdf-java/issues/120

I made the mistake some time back of assuming that the
"scheme" field of AuthScope referred to the URL scheme (e.g. http or https).
But it instead refers to the authentification scheme (e.g. BASIC).
I thought I caught all of those uses in a previous fix, but apparently not.

So this PR fixes all uses I can find of AuthScope.getScheme() to interpret
the scheme properly.


dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
```diff
@@ -0,0 +1,406 @@
...
+    }
+  }
+
+  //////////////////////////////////////////////////
+  // Class variables
+
+  // Order is important; testing reachability is in the order listed
+  static List<Server> SERVERS;
+
+  static {
```
lesserwhirls: Why not use the embedded test server that gretty spins up?

DennisHeimbigner:
Because remotetest is always up, so I can use it with intellij. I have
never successfully brought up gretty under Intellij.bb

lesserwhirls:
This is just a fix for `httpservices`, right? I guess now I'm confused as to why there is a new dap4 test?

In general, I'd really like to avoid pointing to an external server if at all possible, especially one that is referenced directly by its IP address. If having the client and server code split across the netcdf-java and tds repos is causeing some issues, perhaps that's an indication that it would be best if we created a dap4 project on github that held both the client and server (dap4/* from current netcdf-java and tds repos)? Then the artifacts they produce would be included in the netcdf-java and TDS builds, and they could always run their tests in a consistent environment?

At any rate, there are two ways that you can get a server up and running for use with IntelliJ. One is by pointing IntelliJ to a local tomcat installation (can be as simple as the uncompressed Tomcat download with no configuration), and telling IntelliJ which artifact to use with it. The second way is to just have gretty spin up a tomcat in debug mode, and have IntellJ connect to the remote debugging ports exposed by the gretty controlled tomcat. I'd be happy to make another video showing both, if that'd be helpful.

DennisHeimbigner:
The bare IP address is temporary until I can figure out how to do the
self-contained thredds server. Then we can switch to
remotetest.unidata.ucar.edu.

Dont understand this:
>I guess now I'm confused as to why there is a new dap4 test?
That test was added as part of another fix. This fix should only affect
HTTPAuthUtil and HTTPMethod.

The issue of accessing remotetest technically should also affect
the opendap code, so dap4 is not special in any sense.
I am strongly opposed to having a separate dap4 project. We would
need also to build a separate opendap project as well.

> One is by pointing IntelliJ to a local tomcat installation (can be as simple
> as the uncompressed Tomcat download with no configuration), 
This in effect what I am doing, except that the tomcat is on another machine.
Fail to see any distinction.

 >The second way is to just have gretty spin up a tomcat in debug mode, ...
 Sorry, but I find your videos hard to view because of the use of a black
background. In any case this would be helpful especially if accompanied
by some text instructions.

dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
```diff
@@ -0,0 +1,406 @@
...
+    SERVERS = new ArrayList<>();
+    SERVERS.add(new Server("149.165.169.123", "8080"));
```
dopplershift: In no way should this be merged with this IP address here. We are not doing more of this.

DennisHeimbigner: Ok, test removed until we can resolve this.

lesserwhirls:
> > One is by pointing IntelliJ to a local tomcat installation (can be as simple
> > as the uncompressed Tomcat download with no configuration),
> This in effect what I am doing, except that the tomcat is on another machine.
> Fail to see any distinction.

If you point to tomcat on another machine for initial testing because that's better for your development workflow, that's totally fine, but the actual tests that get committed to the codebase should be as self contained as possible, and use the values set in the testUtils modules `ucar.unidata.util.test.TestDir` class.

> > The second way is to just have gretty spin up a tomcat in debug mode, ...
> Sorry, but I find your videos hard to view because of the use of a black
> background. In any case this would be helpful especially if accompanied
> by some text instructions.

Thank you for that feedback - I'm happy to switch out of dark mode for the videos. Just as I did before, I'll include text instructions in the email. Again, I hope the videos can be useful to a larger audience as part of a developers guide at some point.

httpservices/src/main/java/ucar/httpservices/HTTPMethod.java
```diff
@@ -349,12 +349,10 @@ public HttpResponse executeRaw() throws HTTPException {
       setcontent(rb);
       setheaders(rb, this.headers);
       this.lastrequest = buildRequest(rb, this.settings);
-      AuthScope methodscope = HTTPAuthUtil.uriToAuthScope(this.methodurl);
-      AuthScope target = HTTPAuthUtil.authscopeUpgrade(session.getSessionScope(), methodscope);
       // AFAIK, targethost, httpclient, rb, and session
       // contain non-overlapping info => we cannot derive one
       // from any of the others.
-      HttpHost targethost = HTTPAuthUtil.authscopeToHost(target);
+      HttpHost targethost = new HttpHost(this.methodurl.getHost());
```
bickelj: Use the 3-parameter constructor to resolve issue #140

DennisHeimbigner:
I believe bickelj's last fix is correct. But I need to review all uses
of   HTTPAuthUtil.authscopeToHost(target)
to make sure this same mistake is not repeated elsewhere.

httpservices/src/main/java/ucar/httpservices/HTTPMethod.java
rschmunk: Yes. Using HttpHost targethost = new HttpHost(this.methodurl.getHost(), this.methodurl.getPort(), this.methodurl.getScheme()); worked for me.

DennisHeimbigner:
Added an additional minor fix, but this PR should be good to go
modulo jenkins/travis.

lesserwhirls:
Looks like travis is failing on 

~~~java
ucar.nc2.util.net.TestHang > testSession[0] STANDARD_OUT
    Connecting to http://localhost:8080/dts/test.00.dds

ucar.nc2.util.net.TestHang > testSession[0] STANDARD_ERROR
    ucar.httpservices.HTTPException: org.apache.http.conn.HttpHostConnectException: Connect to localhost:80 [localhost/127.0.0.1] failed: Connection refused (Connection refused)
    	at ucar.httpservices.HTTPMethod.executeRaw(HTTPMethod.java:371)
    	at ucar.httpservices.HTTPMethod.execute(HTTPMethod.java:314)
    	at ucar.nc2.util.net.TestHang.testSession(TestHang.java:87)
        ...snip...
    Caused by: org.apache.http.conn.HttpHostConnectException: Connect to localhost:80 [localhost/127.0.0.1] failed: Connection refused (Connection refused)
    	at org.apache.http.impl.conn.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:156)
    	at org.apache.http.impl.conn.PoolingHttpClientConnectionManager.connect(PoolingHttpClientConnectionManager.java:374)
    	at org.apache.http.impl.execchain.MainClientExec.establishRoute(MainClientExec.java:393)
    	at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:236)
    	at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:186)
    	at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:89)
    	at org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110)
    	at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:185)
    	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:72)
    	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:56)
    	at ucar.httpservices.HTTPMethod.executeRaw(HTTPMethod.java:364)
    	... 56 more
    Caused by: java.net.ConnectException: Connection refused (Connection refused)
    	at java.net.PlainSocketImpl.socketConnect(Native Method)
    	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
    	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
    	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
    	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
    	at java.net.Socket.connect(Socket.java:589)
    	at org.apache.http.conn.socket.PlainConnectionSocketFactory.connectSocket(PlainConnectionSocketFactory.java:75)
    	at org.apache.http.impl.conn.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:142)
    	... 66 more

ucar.nc2.util.net.TestHang > testSession[0] STANDARD_OUT
        return from http://localhost:8080/dts/test.00.dds status= 500

ucar.nc2.util.net.TestHang > testSession[0] FAILED
    java.lang.AssertionError: Expected 404: return status: 500
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at ucar.nc2.util.net.TestHang.testSession(TestHang.java:94)
~~~

I made a short video about how to use gretty in Intellij for debugging if that's helpful. I used debugging `ucar.nc2.util.net.TestHang` as an example. The video can be viewed [here](https://drive.google.com/file/d/13HjmPlPHrccMdXQgi0_0oQPkLzqORlwr/view?usp=sharing).

DennisHeimbigner:
so on travis, gretty is not automatically started?

lesserwhirls:
gretty runs automatically when we run our tests. The current failure of `ucar.nc2.util.net.TestHang` on travis is due to the changes in this pull request. I made the video to show you how to debug from inside IntelliJ using gretty to start up the server that the test needs to be able to run, that way you can figure out what in your changes is causing our previous tests to fail now.

rschmunk:
Just checking back on the status of this. In working with the latest 5.3 snapshot the last couple days, I realized I was again running into the problems described in #140 because this PR had not been committed.

DennisHeimbigner:
BTW
apparent;ly the preferred way to do credentials is to
catch requests that fail because of authentication
and then resend the request with preemptive credentials.
This cleans up some of the code and complicates other parts.
It also eliminates the need for asking for password
in the credentials provider.

bickelj:
My use case is unauthenticated, so I cannot speak to the authentication issues. However, is it possible that this integration test is reproducing issue #140?
> Connecting to http://localhost:8080/dts/test.00.dds
followed by
> Connect to localhost:80 [localhost/127.0.0.1] failed: Connection refused (Connection refused)

Does the same integration test failure occur locally?

Does the suggested change to the HttpHost constructor resolve this integration test failure locally?

Should issue #140 be separated from the authentication issues or is it tightly coupled to those issues?

lesserwhirls:
Closing in favor of updated version in PR Unidata/netcdf-java#188

PR #145, opened 2019-11-08
NetcdfFiles and NetcdfDatasets handles both old and new iosps.

JohnLCaron:
Can access both old and new from NetcdfFiles and NetcdfDatasets.
This allows incremental upgrading of iosps

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/145)

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/AttributeContainer.java
+++ b/cdm/core/src/main/java/ucar/nc2/AttributeContainer.java
@@ -6,11 +6,12 @@
 package ucar.nc2;
 
 /**
- * A Container of Attributes.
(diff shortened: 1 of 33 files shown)


lesserwhirls:
It looks like the TDS build is failing at the same spot (`Bearing`) as it was in the previous PR. Do you have the sense that there is more breakage for the TDS with this PR?

JohnLCaron:
ill send a PR for TDS, then we can see.

lesserwhirls:
The TDS tests pass now for this PR - w00t! Looks like just spotless now.

JohnLCaron:
rock and roll

PR #147, opened 2019-11-15
Iosp hdf4

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/147)

### DIFF
--- a/cdm/core/src/main/java/ucar/ma2/DataType.java
+++ b/cdm/core/src/main/java/ucar/ma2/DataType.java
@@ -15,14 +15,9 @@ import java.nio.ByteBuffer;
  * @author john caron
  */
 public enum DataType {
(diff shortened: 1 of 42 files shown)


JohnLCaron:
some tweaks to CompareNetcdf2.ObjFilter causing compile errors in

JohnLCaron:
ready for review

lesserwhirls:
With exception of the "does this method need to be public" questions (and that's just to make sure they do need to be public, as they look like they were package private before), I'm happy to pick up any changes on another PR. I used Reviewable this time, so let's see how that goes.

JohnLCaron:
ready i think

lesserwhirls:
Thanks @JohnLCaron!

PR #153, opened 2019-11-20
Refactor NEXRAD level2 format checking

dopplershift:
Don't try to enumerate all the possible values. Instead, just look for AR2V and a number--and log a warning is this number is bigger than the most recent we are aware of. Also refactor to eliminate duplicated code.

This is motivated by NWS ROC testing data declaring a format of AR2V0008.

@lesserwhirls I'm not having a lot of luck building locally at the moment. Can you double check my work here against the test file?


dopplershift:
I left these in `Level2VolumeScan.java`:
```java
  // data formats
  static final String ARCHIVE2 = "ARCHIVE2";
  static final String AR2V0001 = "AR2V0001";
  static final String AR2V0002 = "AR2V0002";
  static final String AR2V0003 = "AR2V0003";
  static final String AR2V0004 = "AR2V0004";
  static final String AR2V0006 = "AR2V0006";
  static final String AR2V0007 = "AR2V0007";
```
because I'm not sure if they need to be left in the public API. Otherwise, I've eliminated the need for them. (Note the missing AR2V0005, which is a valid option).

lesserwhirls:
See Unidata/netcdf-java#154

PR #155, opened 2019-11-22
HDF5 iosp with builders

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/155)

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/Group.java
+++ b/cdm/core/src/main/java/ucar/nc2/Group.java
@@ -997,6 +997,10 @@ public class Group extends CDMNode implements AttributeContainer {
       return this;
     }
 
+    public Optional<EnumTypedef> findEnumTypedef(String name) {
+      return this.enumTypedefs.stream().filter(e -> e.shortName.equals(name)).findFirst();
+    }
+
     /**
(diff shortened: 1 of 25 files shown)


JohnLCaron:
done.

Note that HdfEos support with HDF5 is not finished.

JohnLCaron:
Note that String.format() uses the %s form, not the {}.

          log.warn(String.format("adjustDimensions %s missing vinfo (new)%n", v.shortName));

No particular reason to prefer over 

  log.warn("adjustDimensions {} missing vinfo (new)", v.shortName);

except that sometimes I like to do printf directly to stdout when debugging. 

Its too bad slf4j didnt use the standard java idiom. I prefer flogger as a logging package, "fluent logging" API. But only works with java.util.logging for the "backend".

rschmunk:
Whoops, @JohnLCaron, sorry about that. Somehow I stared at that bit of code and never noticed the `String.format`.

lesserwhirls:
Flogger looks really cool, and the [Anatomy of a logging API](https://google.github.io/flogger/anatomy) document is an interesting read. Thanks @JohnLCaron !

Issue #149, opened 2019-11-16
Cannot open HDF-EOS swath file with unlimited dimension


rschmunk:
A user reported that Panoply could not open an HDF-EOS file and there was a "Conflicting Dimensions" exception message. The same occurs using toolsUI and IDV. Although the copy of Panoply was using netCDF-Java 5.1, I encountered the same problem when investigating using 5.3-SNAPSHOT.

The exception is thrown at line 262 of `ucar.nc2.iosp.hdf4.HdfEos`.After adding a lot of additional logging msgs, I eventually found that what was going on is that the particular swath file has metadata stating that the time dimension nTime has length 1, like so:

```
	GROUP=SWATH_1
		SwathName="MOP02"
		GROUP=Dimension
			OBJECT=Dimension_1
				DimensionName="Unlim"
				Size=-1
			END_OBJECT=Dimension_1
			OBJECT=Dimension_2
				DimensionName="nTime"
				Size=1
			END_OBJECT=Dimension_2

```

However, the time dimension actually has 221967 steps, and the various variables that use that dimension specify a MaxdimList of unlimited, e.g.,

```
			OBJECT=DataField_2
				DataFieldName="SolarZenithAngle"
				DataType=H5T_NATIVE_FLOAT
				DimList=("nTime")
				MaxdimList=("Unlim")
			END_OBJECT=DataField_2
```

So it seems that on trying to acquire the dataset, the netCDF-Java library is figuring out that there are 221967 timesteps in the example file, but when it circles around and does some testing to verify that this is an HDF-EOS file, it's running into that bit of metadata that says the nTime dimension has size 1. And thus the conflicting dimensions exception gets thrown.

I commented out that exception throw, and was then able to open the file and plot from the data within with problem. However, I don't know that that's  the best solution for this problem. 🙄

For an example dataset, I ran my tests using ftp://l5ftl01.larc.nasa.gov/pub/MOPITT/MOP02J.008/2017.08.17/MOP02J-20170807-L2V18.0.3.he5

JohnLCaron:
I just ported HDF-EOS to new 5.3 framework. Would be happy to test if we can get an example file.

rschmunk:
@JohnLCaron, see ftp URL given at the end of my initial post.

JohnLCaron:
[...]

PR #162, opened 2019-12-06
HdfEos dimension conflicts

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/162)

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/Dimensions.java
+++ b/cdm/core/src/main/java/ucar/nc2/Dimensions.java
@@ -1,6 +1,7 @@
 /* Copyright Unidata */
 package ucar.nc2;
 
+import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.Formatter;
 import java.util.List;
@@ -84,14 +85,14 @@ public class Dimensions {
    * @return equivalent list of Dimension objects.
    * @throws IllegalArgumentException if cant find or parse the name.
    */
-  public static List<Dimension> makeDimensionsList(Find finder, String dimString) throws IllegalArgumentException {
-    List<Dimension> newDimensions = new ArrayList<>();
+  public static ImmutableList<Dimension> makeDimensionsList(Find finder, String dimString) throws IllegalArgumentException {
     if (dimString == null) // scalar
-      return newDimensions; // empty list
+      return ImmutableList.of(); // empty list
     dimString = dimString.trim();
     if (dimString.isEmpty()) // scalar
-      return newDimensions; // empty list
+      return ImmutableList.of(); // empty list
 
+    ImmutableList.Builder<Dimension> newDimensions = ImmutableList.builder();
     StringTokenizer stoke = new StringTokenizer(dimString);
     while (stoke.hasMoreTokens()) {
       String dimName = stoke.nextToken();
@@ -114,7 +115,7 @@ public class Dimensions {
       newDimensions.add(d);
     }
 
-    return newDimensions;
+    return newDimensions.build();
   }
 
   /** Make a list of private dimensions from an array of lengths */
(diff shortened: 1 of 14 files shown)


JohnLCaron:
Issue #149

JohnLCaron:
https://artifacts.unidata.ucar.edu/repository/unidata-releases/org/bounce/bounce/0.18/bounce-0.18.pom

artifacts is down??

dopplershift:
I can get to https://artifacts.unidata.ucar.edu just fine, but not to that URL...

lesserwhirls:
The issue isn't that the bounce does not exist (that's just gradle checking the repos for the artifact - it actually lives in our 3rd party repo), the issue is that gradle cannot do a HEAD request due to not being able to understand our certificate. This is the key error is:

~~~java
Caused by: java.lang.RuntimeException: Could not generate DH keypair
	... 182 more
Caused by: java.security.InvalidAlgorithmParameterException: DH key size must be multiple of 64, and can only range from 512 to 2048 (inclusive). The specific key size 4096 is not supported
~~~

@oxelson - did we recently change the cert on artifacts? This error was generated from Oracle JDK 1.8 u151 as used on the Travis CI service, and we see similar issues on Jenkins. It looks like java 8 only support 2048 for Diffie-Hellman. Java 9 would give us 3072, 4096, 6144, 8192.

There is a [reported work around](https://stackoverflow.com/questions/6851461/why-does-ssl-handshake-give-could-not-generate-dh-keypair-exception/29176597#29176597), but man...

imma head out

oxelson:
@lesserwhirls yes, we modified our ciphers to be compatible with TLS 1.3.  Unfortunately, it looks like the Sonatype Nexus Repository Manager currently doesn't support OpenJDK versions greater than version 8 (https://issues.sonatype.org/browse/NEXUS-19183).

I can back us out of TLS 1.3 on that server for now, but it needs to be considered a temporary fix.  I'll also add our voice to the chorus on the Sonatype issue tracker for support of OpenJDK 11.

oxelson:
@lesserwhirls, @JohnLCaron  TLS1.3 has been disabled.

lesserwhirls:
Thanks Jen! I'm surprised this is an issue related to java on the server side, but I totally believe it. The message above was from a client-side stacktrace, as in it was the JVM client-side that didn't understand what to do with 4096 bit DH cert it was handed by the server. I was able to get to the server through my browser, and the browser showed it was using the 4096 bit DH cert, so I assume it was all a client-side issue.

Now that we can grab artifacts and build, we're seeing the same thing in our tests when they try to hit thredds-dev, for example:

https://travis-ci.com/Unidata/netcdf-java/jobs/263696653#L16900-L16952

oxelson:
Yeah, it's a client issue and we're running into it all over the place.  :-(   I enabled all of our servers to use TLS 1.3.  I'll roll back thredds-dev and thedds-test.  Do I need to do so for thredds.ucar.edu?

lesserwhirls:
That would be great if you could with the total understand on my end that it's temporary. AdoptOpenJDK/openjdk-build#1254 looks promising. Azul JREs/JDKs have it now (I wonder if nexus could run under Azul's [Zulu Community JDK](https://www.azul.com/downloads/zulu-community/?&architecture=x86-64-bit&package=jdk)). Oracle plans on adding TLS 1.3 support to their JRE/JDK 8 [sometime in 2020](https://java.com/en/jre-jdk-cryptoroadmap.html).

Maybe there is some hope for Java 8 (although we need to, and will move ourselves), specifically for others who are "stuck" on 8 but can still use the latest Java 8.

oxelson:
giphy



Okay, TLS 1.3 has been disabled for the thredds-dev, thredds-test and thredds hosts.  


With modern browsers to stop supporting ciphers < TLS 1.2 next year, I suspect (hope) we're going to see a scramble to adopt clients that support TLS 1.3 ciphers (including java).

lesserwhirls:
Ok, we're green on this one now - merging! Thanks @JohnLCaron!

PR #163, opened 2019-12-06
Stream iosp builder

JohnLCaron:
Add builder for NcStreamIosp

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/163)

PR #164, opened 2019-12-07
Grib iosp builder

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/164)

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/FileWriter2.java
+++ b/cdm/core/src/main/java/ucar/nc2/FileWriter2.java
@@ -26,9 +26,6 @@ import java.util.Map;
  * <p>
  * Use a NetcdfFileWriter object for a lower level API.
  *
- * @see ucar.nc2.dt.grid.CFGridWriter2
- * @see ucar.nc2.ft.point.writer.CFPointWriter
- *
  * @deprecated use ucar.nc2.writer.FileWriter
  */
 public class FileWriter2 {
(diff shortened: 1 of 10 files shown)


lesserwhirls:
Just a minor question - will pick up Unidata/netcdf-java#163 with this one since it's included.

JohnLCaron:
NcMLReader.writeNcMLToFile seems wrong (writing from a reader class). Its a static method, it could go anywhere.

Not sure who calls it (TDS?) . Im thinking maybe we should move it somewhere else (NcmlWriter?)

What do you think?

lesserwhirls:
The TDS codebase does not call it. A google search didn't turn up anything useful. I don't really see another method for reading NcML from an input stream and writing out a netCDF file to disk, and we do use it in ToolsUI to read text from the NcML editor and write it to a netCDF file:

https://github.com/Unidata/netcdf-java/blob/270500e8f1d5d35e1b19e8890e3c1ec2662a0641/uicdm/src/main/java/ucar/nc2/ui/op/NcmlEditor.java#L298-L311

The current `NcmlFileWriter` seems to be focused on writing actual NcML, but I can't think of a better place. We could stuff the logic inside the `NcmlEditor.writeNetcdf` method since there isn't much involved (only five lines), but that does not feel 100% right either.

JohnLCaron:
I think we should eliminate unneeded methods. We can put the code in NcmlEditor if needed.

Would be good to get this PR in, as I have another one ready to go.

lesserwhirls:
Sounds good to me. Merging now. Thanks @JohnLCaron!

PR #169, opened 2019-12-26
TdsMonitor fixes and improvement

lesserwhirls:
This PR fixes a few bugs, namely:

1. Y-Axis labels on the AccessLogs panel's TimeSeries chart:

   **Before:**
   before_chart

   **After:**
   after_chart

1. Minor clarification of the TimeSeries chart's x-axis label.

1. Allow logs that exists on disk, but have not been downloaded by TdsMonitor directly, to be analyzed (initilized by typing the name of the server into the `server` text box in the ManageLogs panel).

This PR also introduces an improvement, specifically by allowing users to choose date/time limits on the access and servlet logs via a [LGoodDatePicker](https://github.com/LGoodDatePicker/LGoodDatePicker/) DateTimePicker widget rather than editing an ISO time string by hand.

* **Before:**
   before_dateTime

* **After:**
   after_dateTime


lesserwhirls:
Happy New Year `TdsMonitor`!

PR #171, opened 2019-12-26
Attribute containers, API cleanup.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/171)

### DIFF
--- a/bufr/src/main/java/ucar/nc2/ft/point/bufr/StandardFields.java
+++ b/bufr/src/main/java/ucar/nc2/ft/point/bufr/StandardFields.java
@@ -215,12 +215,12 @@ public class StandardFields {
       private Field(TypeAndOrder tao, Variable v) {
         this.tao = tao;
         this.memberName = v.getShortName();
-        Attribute att = v.findAttribute("scale_factor");
+        Attribute att = v.attributes().findAttribute("scale_factor");
         if (att != null && !att.isString()) {
           scale = att.getNumericValue().doubleValue();
           hasScale = true;
         }
-        att = v.findAttribute("add_offset");
+        att = v.attributes().findAttribute("add_offset");
         if (att != null && !att.isString()) {
           offset = att.getNumericValue().doubleValue();
           hasScale = true;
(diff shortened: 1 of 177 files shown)


JohnLCaron:
Youll need to approve the other one, then i will sync, before this ones ready to go.

JohnLCaron:
Not clear whats failing.

lesserwhirls:
I reran the test and things passed. That particular failure reminded me an awful lot of what https://github.com/Unidata/netcdf-java#61 fixed, at least from what I've seen on Jenkins. Maybe we have another race condition? I've lost the output from the failing test, but I think it was in the same place ("InterupptedException=sleep interupted" during TestNetcdfFileCache > testPeriodicClear). At any rate, something to keep an eye out for.

Given that these commits are on https://github.com/Unidata/netcdf-java/pull/172, can you rebase that one and we'll get these in over there?

JohnLCaron:
Only some of the changes on this one are on #172, so we have to get this one in first.
I just rebased to master, so this one is ready as soon as it passes travis.

I should note that I was trying to finish all attribute related changes, but I still need to review NetcdfFile "version 3 compatibility" interface.

lesserwhirls:
Looks good to me. Having looked this over though, I'm wondering if some of the attribute methods should use `Optional<>`, like `getUnitsString()`. Or maybe the `@Nullable` and `@NotNull` annotations give a good enough suggestion? At any rate, merging this one now.

JohnLCaron:
Optional is a good (probable better) alternative to returning null. However each place its used has to use a rather different syntax that takes some getting used to. Also, it has to be renamed from the old method. So I find it cumbersome to retrofit, and also to mix the two. So tentatively Ive decided not to use it. Open to persuation.

JohnLCaron:
Id like to consistently get everything marked @Nullable. The idea is to not use @Nonnull, if its not marked @Nullable then it cant be null. 
Also waiting for a new standard annotations package, see:

Replace jsr305 (standard annotations) with https://docs.google.com/document/d/1d5NgP4Pjk_ahkISpRa3nFb1FzPz9yS9luByiBj6qCsk

JohnLCaron:
Sorry I thought that doc was public. Ill get a copy.

Issue #22, opened 2019-07-16
Support GRIB GDT 140 and PDT 73


msdsoftware:
A user sent me a GRIB2 file of ECMWF flood data that netCDF-Java will not open.

The first problem I encountered in trying to figure out why e.g. Panoply and IDV would not open the file is that it specified Grid Definition 140, which is the Lambert Azimuthal Equal Area projection. Some Googling indicates that this projection was first proposed for addition to GRIB in 2012, which I expect is after most/all of the grids that NJ understands were coded.

After an attempt at hacking `Grib2Gds` to accept template 140, I then ran into the problem that the file uses Product Definition 73, which is missing from `Grib2Pds`. 

Perhaps there are further problems, but that was where I quit.

msdsoftware:
The dataset creator has suggested they might switch to a different Product Definition Template. He didn't say which, but I would think it would be likely one of the other forecast templates 70-72.

lesserwhirls:
Thanks for the report, @msdsoftware! I'll take a look once we get 5.0.0 stable out the door (likely today). Do you have a sample file I could work with? Even if they switch templates, it'd still be nice to take a look and see if we can support it.

msdsoftware:
@lesserwhirls, I have a couple sample datasets. Would have to double-check but I think one should be okay for redistribution.

cofinoa:
@msdsoftware please can you share the sample datasets?

ms-daniel-bailey:
Hi guys, hope it's appropriate to piggyback this issue as it seems relevant and I have a couple of example files I can share for reference sake.

I'd been wrestling with reading these files with software I am working on, despite the software being stable until encountering these particular grid data types.   I ended up deciding to try loading them in Panoply which has also crashed and burned with the same NullException I have been experiencing - which comes as a result of Grib2PDS not handling product definition 32, which is required to handle these files.

If there are any suggestions of how to work around this, that would be appreciated - but otherwise this seems like something which should be able to work as it appears that this product definition is known.

Cheers,
Dan
[ACCESS-G-APS3-MEDIA_2019122212_000_95.228.zip](https://github.com/Unidata/netcdf-java/files/4050514/ACCESS-G-APS3-MEDIA_2019122212_000_95.228.zip)

rschmunk:
FWIW, the issue of Product Definition 32 support arose two years ago in Unidata/thredds#846. Looks like @cofinoa took a whack at it but it turned out to be pretty complex and so apparently was not completed.

PR #182, opened 2020-01-13
Grib2Pds throw UnsupportedOperation rather than return null

rschmunk:
Grib2Pds returns a null if product template is not recognized. Higher up the chain, this results in an unhelpful NullPointerException being thrown. Better to throw an UnsupportedOperation exception with a useful message. Note that the latter way is how Grib2Gds handles an unrecognized grid template.

### DIFF
--- a/grib/src/main/java/ucar/nc2/grib/grib2/Grib2Pds.java
+++ b/grib/src/main/java/ucar/nc2/grib/grib2/Grib2Pds.java
@@ -68,7 +68,7 @@ public abstract class Grib2Pds {
         return new Grib2Pds61(input);
       default:
         log.warn("Missing template " + template);
-        return null;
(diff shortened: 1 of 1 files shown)


lesserwhirls:
Good call - this will help make things more clear when we don't understand the PDS. Thank you @rschmunk! I'll apply the code style so that the last check passes.

rschmunk:
TY, @lesserwhirls.

rschmunk:
TBH, this was in response after taking a look at #22.

rschmunk:
One clarification. If Grib2Gds does not recognize the template, the thrown exception gets caught in Grib2CollectionBuilder so that the user may instead see an IOException stating "No records found in dataset".

But if this new exception in Grib2Pds gets thrown, it comes all the way out where the user can see it.

Issue #185, opened 2020-01-15
DatasetUrl.findDatasetUrl(location) fails with https://upwell.pfeg.noaa.gov/erddap/griddap/jplAquariusSSS3MonthV5


noaaroland:
Version 5.2.0 with openjdk version "1.8.0_212" on Ubuntu 16.4

When opening the above OPeNDAP URL with 5.2 the findDatasetUrl fails with a 404 not found.

When opening the same URL with 4.6.14 the code goes straight to the:
 NetcdfDataset ncd = NetcdfDataset.acquireDataset(location, task);

and the data set opens just fine.

rschmunk:
The dataset can be loaded using NJ 5.1, but fails with a 5.3 snapshot.

lesserwhirls:
Fixed by Unidata/netcdf-java#188 - thank you for the report @noaaroland and additional info @rschmunk !

PR #188, opened 2020-01-20
Fix use of AuthScope.getScheme() - modified PR

lesserwhirls:
Based on PR Unidata/netcdf-java/#141, with special thanks to @bickelj (https://github.com/Unidata/netcdf-java/pull/141#issuecomment-573038077)

Should fix Unidata/netcdf-java#140, Unidata/netcdf-java#120, and Unidata/netcdf-java#185.


lesserwhirls:
Ok, so not Unidata/netcdf-java#120, but I think we're in better shape getting this in once everything is green.

lesserwhirls:
Spotless failed, but that particular failure is fixed in another PR. Otherwise, we’re good on this one, so merging.

bickelj:
> Based on PR Unidata/netcdf-java/#141, with special thanks to @bickelj ([#141 (comment)](https://github.com/Unidata/netcdf-java/pull/141#issuecomment-573038077))

Yw, but thank you for making the change, testing it, merging it, etc. I look forward to the next release!

Issue #174, opened 2019-12-31
Nexrad2IOServiceProvider.isNEXRAD2Format() failing on hi res


JohnLCaron:
https://jenkins-aws.unidata.ucar.edu/job/netcdf-java/lastCompletedBuild/testReport/ucar.nc2.iosp.nexrad2/TestNexrad2HiResolution/testRead/

java.io.IOException: java.lang.NumberFormatException: For input string: "IVE2"
	at ucar.nc2.NetcdfFile.open(NetcdfFile.java:366)
	at ucar.nc2.dataset.NetcdfDataset.openProtocolOrFile(NetcdfDataset.java:814)
	at ucar.nc2.dataset.NetcdfDataset.openFile(NetcdfDataset.java:673)
	at ucar.nc2.iosp.nexrad2.TestNexrad2HiResolution$MyAct.doAct(TestNexrad2HiResolution.java:54)
	at ucar.unidata.util.test.TestDir.actOnAll(TestDir.java:299)
	at ucar.unidata.util.test.TestDir.actOnAll(TestDir.java:243)
	at ucar.nc2.iosp.nexrad2.TestNexrad2HiResolution.testRead(TestNexrad2HiResolution.java:30)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

appears to be from Ryan's change on 11/20 at revision:

54a554c1d2a32e8eb939b8ce7ce4c3a064b4e8fa

lesserwhirls:
I have a fix for this on in my S3 PR (one of my new tests barfed on it). Should be merged in today (just finishing up a service provider for `RandomAccessFile`).

JohnLCaron:
Still failing in

https://jenkins-aws.unidata.ucar.edu/job/netcdf-java/lastCompletedBuild/testReport/

PR #173, opened 2019-12-30
S3 Support

lesserwhirls:
Add support for reading S3 objects as RandomAccessFiles, based off of the work done at https://github.com/Unidata/thredds/tree/feature/s3+hdfs

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/173)

PR #189, opened 2020-01-21
Fix ARV2 version check

lesserwhirls:
Fixes Unidata/netcdf-java#174 

Pulled from Unidata/netcdf-java#173, as I'm going to look at added a `LoadingCache` for S3 and HTTP remote random access files, but I don't want to hold this fix up.


JohnLCaron:
Alright, jenkins has no failures!

lesserwhirls:
Don't worry... 

wreck it

PR #198, opened 2020-02-03
Demonstrate use of NetcdfFile builder in javadoc

rschmunk:
Javadoc for NetcdfFile class showed use of open method for accessing file, but that method is deprecated.

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/NetcdfFile.java
+++ b/cdm/core/src/main/java/ucar/nc2/NetcdfFile.java
@@ -56,7 +56,7 @@ import ucar.unidata.util.StringUtil2;
  * <pre>
  * NetcdfFile ncfile = null;
  * try {
- *   ncfile = NetcdfFile.open(fileName);
+ *   ncfile = NetcdfFile.builder ( ).setLocation (fileName).build ( );
  *   // do stuff
  * } finally {
  *   if (ncfile != null) {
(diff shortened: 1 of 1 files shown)


lesserwhirls:
Greetings @rschmunk! Thank you for trying to improve the javadocs! The new builder interface isn't quite ready to use, which is why it has not been documented yet (even though the deprecation messages are screaming on build...sorry about that). The builder will be used to build up a NetcdfFile object, and is used by the new IOSPs quite a bit. However, I suspect most people will be trying to open an existing file. This can be done using the `open(...)` static helper methods from `ucar.nc2.NetcdfFiles`. Once I get `5.3` out the door, I plan on writing up a blog post about where things are heading, what the plan is for minimum JDK support, when things marked deprecated will disappear, etc.

rschmunk:
Looking forward to the blog post, @lesserwhirls.  I've shifted over to using 5.3 SNAPSHOT, but the number of deprecation notices that I'm getting is more than irritating. I've managed to knock it down from 68 to 42 so far. Could knock out another two dozen if I could figure out what to do about `ProjectionImpl`.

rschmunk:
Okay, I give up. Despite my best of intentions, spotless is going to fail my changes to javadoc.

rschmunk:
Okay, I give up. I can try to update javadoc, and spotless will get hyper-sensitive and deny it. Please mark this PR as rejected.

lesserwhirls:
I appreciate your efforts, and I certainly don't mean for spotless to be overbearing. Were you able to run `./gradlew spotlessApply` locally? That should apply the style - if you did that and the spotless check on Travis fails, then something terrible is happening in-between.

The projection stuff is a work in progress. @JohnLCaron - any guidance on `ProjectionImpl`?

rschmunk:
My mistake was in trying to edit the javadoc directly in the Unidata repo on the Github website. I should have worked on a local copy of my fork where I could run gradlew and have it apply splitless for me.

I've not run into spotless before, so had no idea that a seemingly simple javadoc edit could run into this sort of trouble.

PR #203, opened 2020-02-07
Update NetcdfFile header javadoc

rschmunk:
This replaces #198 which was failing due to a spotless conflict.

Demonstrates use of `NetcdfFile.builder()`, but then recommends use of `NetcdfFiles` statics.


lesserwhirls:
Thanks @rschmunk! The only thing I would change here is that the builder does not actually open a file, but rather is used to build a `NetcdfFile` object. The sample code you have in the javadoc will build a `NetcdfFile` object with a location set to `fileName`, but it does not actually try to use a file located at `fileName` to prepopulate the object.

rschmunk:
Reworded.

As an FYI, I got into this while updating my own code to deal with deprecations in NJ, but seeing that the javadoc for `NetcdfFile` and `NetcdfDataset` still showed sample code using the deprecated way.

I also revised, e.g., my use of `CoordinateAxis1D`, but the javadoc for that class was not in conflict with 5.3 changes.

lesserwhirls:
Thank you @rschmunk!

PR #209, opened 2020-02-17
BUFR TableD Fixes

lesserwhirls:
Not every sequence entry in the WMO xml table D files is processed, only those that contain a specific child element (ElementName_E). For older WMO tables, sometimes that element isn't used, and so the sequence is skipped when reading in the table. However, there are cases where we should probably be looking for another element to use in addition to the one we currently look for. An example of this is `<ExistingElementName_E>`, and we do see entries where `<ElementName_E>` is not used, but `<ExistingElementName_E>` is. Skipping these entries has caused errors, so I'm pretty sure we need to keep those. There are potential sequence entries we skip which do not use either one of those element names, but I'm not sure what to do with them quite yet. They might be cruft describing table headings from the literal tables used in the word and pdf documents, but they have `<FXY1>` and `<FXY2>`, so maybe they should not be skipped? The documentation just isn't there, and the xsd does not say much about anything (`minOccurs="0"` for everybody).

Using WMO v16 Table D as an example, with this change we go from using 4886 potential sequences to using 5294 (out of 5321). The only child elements that appear in every entry are  `<Category>`, `<CategoryOfSequences_E>`, `<FXY1>`, `<FXY2>`, `<No>`, and `<Status>`.

As part of this, I created a utility for making sense of WMO TableD xml files (which allowed me to write the previous paragraph). This utility will read in an xml based WMO table, describe the various ways the sequence entries are defined, and tell you how many entries will be missed by the current netCDF-Java table reader (bufr/src/main/java/ucar/nc2/iosp/bufr/tables/WmoXmlReader.java). I probably had way too much fun with streams doing that.

From what I can tell, starting with v19 of WMO table D xml files, every entry has an `<ElementName_E>`, so that's nice. Not quite the case prior to that, however.

A sample file that drove all of this is available in the test datasets (`cdmUnitTest/formats/bufr/userExamples/WMO_v16_3-10-61.bufr`) and gets checked by `ucar.nc2.iosp.bufr.TestBufrRead` in `cdm-test`.

### DIFF
--- a/bufr/src/main/java/ucar/nc2/iosp/bufr/tables/TableC.java
+++ b/bufr/src/main/java/ucar/nc2/iosp/bufr/tables/TableC.java
@@ -20,6 +20,7 @@ public class TableC {
     tableCdesc[4] = "add associated field";
     tableCdesc[5] = "signify character";
     tableCdesc[6] = "signify data width for next descriptor";
+    tableCdesc[7] = "increase scale, reference value, and data width";
     tableCdesc[21] = "data not present";
     tableCdesc[22] = "quality information follows";
     tableCdesc[23] = "substituted values operator";
--- a/bufr/src/main/java/ucar/nc2/iosp/bufr/BufrIosp2.java
+++ b/bufr/src/main/java/ucar/nc2/iosp/bufr/BufrIosp2.java
@@ -1,18 +1,30 @@
 /*
- * Copyright (c) 1998-2018 University Corporation for Atmospheric Research/Unidata
+ * Copyright (c) 1998-2020 University Corporation for Atmospheric Research/Unidata
  * See LICENSE for license information.
  */
 package ucar.nc2.iosp.bufr;
 
+import java.io.IOException;
+import java.util.Formatter;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
 import org.jdom2.Element;
+import ucar.ma2.Array;
+import ucar.ma2.ArraySequence;
+import ucar.ma2.ArrayStructure;
+import ucar.ma2.Section;
+import ucar.ma2.StructureData;
+import ucar.ma2.StructureDataIterator;
+import ucar.nc2.Group;
+import ucar.nc2.NetcdfFile;
+import ucar.nc2.Sequence;
+import ucar.nc2.Structure;
+import ucar.nc2.Variable;
 import ucar.nc2.constants.DataFormatType;
-import ucar.ma2.*;
-import ucar.nc2.*;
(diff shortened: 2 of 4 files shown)


bufr/src/main/java/ucar/nc2/iosp/bufr/tables/TableC.java
```diff
@@ -20,6 +20,7 @@
     tableCdesc[4] = "add associated field";
     tableCdesc[5] = "signify character";
     tableCdesc[6] = "signify data width for next descriptor";
+    tableCdesc[7] = "increase scale, reference value, and data width";
```
JohnLCaron: did you find what this means? I dont see an implementation in DataDescriptorTreeConstructor??

JohnLCaron:
This looks great, I assume it fixes WMO_v16_3-10-61.bufr?

As soon as its checked in, I have a few BUFR related changes.

bufr/src/main/java/ucar/nc2/iosp/bufr/tables/TableC.java
lesserwhirls: It was in there already, just not defined in `TableC.java`

https://github.com/Unidata/netcdf-java/blob/90ae49b86ba094706af504401f1abe0c7db1f418/bufr/src/main/java/ucar/nc2/iosp/bufr/DataDescriptorTreeConstructor.java#L330

Once `changeWtf` get's set, it is applied here until it gets turned off:

https://github.com/Unidata/netcdf-java/blob/90ae49b86ba094706af504401f1abe0c7db1f418/bufr/src/main/java/ucar/nc2/iosp/bufr/DataDescriptorTreeConstructor.java#L410-L423

lesserwhirls:
> This looks great, I assume it fixes WMO_v16_3-10-61.bufr?

Yep, this fixes the two bufr files `tmp1.bufr` and `tmp2.bufr` the user shared (I renamed `tmp2.bufr` to `WMO_v16_3-10-61.bufr`).

JohnLCaron:
thanks! good work!!
moral of the story: if your town consists entirely of glass houses, you
probably live in one too.

lesserwhirls:
Hey man, you still have windows...my house was glass but it's all shattered at this point!

lesserwhirls:
> Hey man, you still have windows...my house was glass but it's all shattered at this point!

Actually, maybe I misunderstood your statement, and if so I apologize. The town of BUFR is indeed a fragile beast.

JohnLCaron:
I was referring to the house of netcdf-java in the village of scientific data providers and software. So I think you understood.

PR #212, opened 2020-02-24
Squashing a few BUFR bugs

lesserwhirls:
We were identifying unit values `CCITT IA5` and `CCITT_IA5` as indicating descriptor type string, but not `CCITTIA5`. This was causing some tests to fail with the new BUFR files added to thredds-test-data. Now we do this check in one method. This PR also makes sure we do the same forCode Table and Flag table units (also checking for `CODETABLE` and `FLAGTABLE`, as that is what is used in the eumetsat tables). Oh, BUFR...you so silly.

### DIFF
--- a/bufr/src/main/java/ucar/nc2/iosp/bufr/BufrIdentificationSection.java
+++ b/bufr/src/main/java/ucar/nc2/iosp/bufr/BufrIdentificationSection.java
@@ -169,7 +169,9 @@ public class BufrIdentificationSection {
 
       // Optional section octet 10
       int optional = raf.read();
-      hasOptionalSection = (optional & 0x40) != 0;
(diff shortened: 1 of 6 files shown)


lesserwhirls:
A few more BUFR commits here. So, first off, we were not checking the most significant bit correctly to determine if the BUFR message contained the optional section. Second, ECMWF uses the WMO tables at least sometimes, so I added an override to go back to WMO for the one case I know about. Might not be the right approach for tablelookup.csv, and might need to be more specific about what we override, as opposed to being specific about which ECMWF message we allow to fall back to WMO.

lesserwhirls:
Running on Jenkins now, will merge if everything looks good there.

PR #214, opened 2020-02-25
GitHub actions

lesserwhirls:
Enable GitHub Actions for CI and fix a bug uncovered by the transition.

### DIFF
--- a/.github/workflows/netcdf-java.yml
+++ b/.github/workflows/netcdf-java.yml
@@ -1,0 +1,81 @@
+on: [pull_request]
+
+jobs:
+  tests-zulu:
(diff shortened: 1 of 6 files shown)


lesserwhirls:
This will enable GitHub Actions to run the unit tests currently run by Travis. There are several advantages to this:

1. We have access to the JUnit html results as an artifact (zip) attached to each job (available for 90 days). I only have these being produced on failures at the moment.
1. Now running on the latest Ubuntu (as opposed to 16.04)
1. Pulling netCDF-C from the Ubuntu repos as opposed to our stale lib checked into GitHub

PR #231, opened 2020-03-14
deprecate lots of stuff in ucar.nc2.util

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/231)

### DIFF
--- a/cdm/core/src/main/java/thredds/client/catalog/ServiceType.java
+++ b/cdm/core/src/main/java/thredds/client/catalog/ServiceType.java
@@ -83,7 +83,7 @@ public enum ServiceType {
   public enum AccessType {
     Catalog("Catalog"), Metadata("Metadata"), DataAccess("Data Access"), Unknown("Unknown");
 
-    protected String name;
+    protected final String name;
 
     AccessType(String name) {
       this.name = name;
--- a/cdm/core/src/main/java/thredds/client/catalog/tools/CatalogXmlWriter.java
+++ b/cdm/core/src/main/java/thredds/client/catalog/tools/CatalogXmlWriter.java
@@ -18,6 +18,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Iterator;
 import java.util.List;
+import ucar.unidata.util.Format;
 
 /**
  * Write client side catalogs out as XML.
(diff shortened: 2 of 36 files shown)


JohnLCaron:
@lesserwhirls  can you look at check failures, not sure how to interpret.

lesserwhirls:
Ok. So, first, sorry about the delay. F. Second, there appear to be a few things happening. One is that github actions barfed when trying to install netCDF-C using `apt` in Ubuntu. However, if I look at the travis build, at least one of the tests fails to compile because of a change made to a method access modifier in `DatasetUrl.java`:

```java
/home/travis/build/Unidata/netcdf-java/cdm/core/src/test/java/ucar/nc2/dataset/TestDatasetUrl.java:30: error: getProtocols(String) has private access in DatasetUrl
```

Might be able to make `DatasetUrl.getProtocols(...)` package-private instead of private if we want to keep the test. I'm re-running the GitHub Actions builds, and I see we are now failing in the same spot (looks like the previous unrelated issue with `apt` is resolved).

JohnLCaron:
ready

lesserwhirls (approved):
> Reviewed 36 of 36 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/231)* status: 35 of 36 files reviewed, 1 unresolved discussion (waiting on @JohnLCaron and @lesserwhirls)

___
*[cdm/core/src/main/java/ucar/nc2/util/Optional.java, line 10 at r1](https://reviewable.io/reviews/unidata/netcdf-java/231#-M2nmJOv3Ttya9z5vYZX:-M2nmJOv3Ttya9z5vYZY:b-mqkfhr) ([raw file](https://github.com/unidata/netcdf-java/blob/626c72813c652e79dd74b8fb3b45e415d8d795bd/cdm/core/src/main/java/ucar/nc2/util/Optional.java#L10)):*
> ```Java
>  * Optional with an error message when empty
>  * 
>  * @deprecated will move to java.util.Optional in ver6
> ```

The message aspect seems to be a pretty nice addition. Is there a clean way of doing that with the vanilla Optional?

JohnLCaron:
> The message aspect seems to be a pretty nice addition. Is there a clean way of doing that with the vanilla Optional?

Good question, the answer is no. Looking at the way it was actually used, I decided that it wasnt useful enough to justify a non-standard implementation of Optional. However, its used ~200 times in ft2, which argues to let it be an internal class. Mostly I want to deprecate it for any further use, and get it out of the public API.

lesserwhirls:
Sounds good to me - thanks @JohnLCaron !

PR #247, opened 2020-04-10
First pass at correcting know issues with URLs in links and references.

oxelson:
In netcdf-java repo:

- removed unidata downloads site index.jsp references
- removed references to my.unidata domain
- removed references to old package convention in unidata urls


lesserwhirls:
Thanks @oxelson!

PR #248, opened 2020-04-10
http --> https

oxelson:
correction of http protocol in relevant places.


oxelson:
@lesserwhirls  Otay... netcdf-java http --> https stuff.

lesserwhirls:
Man those `@see` links in the javadocs that point to documentation in the userguide. They need to be updated for sure. We should probably add a custom tag to our javadocs to generate the correct versioned link for those types of things, too. I'll make an issue for those tasks and will circle around to them later. Thanks @oxelson!

PR #254, opened 2020-04-14
update cdm-mcidas

lesserwhirls:
Update the source files in `ucar.mcidas` with the latest and greatest (I think...) from the `edu.wisc.ssec.mcidas` package from visad.

Also, tell spotless to ignore these files. We'll keep those as close to the original as possible.


lesserwhirls:
whoa...this is a bad place to live...closing it out

PR #259, opened 2020-04-17
Add a version string to the build configuration

ddirks:
Add a version string to the build configuration, and reflect that version number in a meta tag in the head of the built HTML files.


ddirks:
Hi @lesserwhirls -- trying to get everything into a known state. I'll have more changes but would like to get this PR merged before I continue.

ddirks:
@lesserwhirls now it appears some of John's commits are included here? 🤯

PR #268, opened 2020-04-22
Add doc version menu

ddirks:
This adds a dynamic version menu to the netcdf-java docs, allowing the user to switch to the current topic in the docset for a different listed version.

Comments especially solicited on robustness/error-checking in the javascript (customscripts.js), behavior in "responsive" situations (small screens), and visual presentation.

### DIFF
--- a/docs/src/public/userguide/_config.yml
+++ b/docs/src/public/userguide/_config.yml
@@ -3,6 +3,7 @@ output: web
 # this property is useful for conditional filtering of content that is separate from the PDF.
 
 docset_version: 5.4
+# this will appear in an HTML meta tag, sidebar, and perhaps elsewhere
 
 topnav_title: netCDF-Java Documentation
 # this appears on the top navigation bar next to the home button
--- a/docs/src/public/userguide/_config.yml
+++ b/docs/src/public/userguide/_config.yml
@@ -101,6 +101,7 @@ description: "This is a documentation site for the netCDF-Java and THREDDS Data 
 # the description is used in the feed.xml file
 
 # needed for sitemap.xml file only
-url: www.unidata.ucar.edu
+# must include trailing slash, leave off the version :-)
+base_docs_url: https://docs.unidata.ucar.edu/netcdf-java/
 
 google_analytics: UA-92978945-1
--- a/docs/src/public/userguide/_config.yml
+++ b/docs/src/public/userguide/_config.yml
@@ -17,13 +17,7 @@ site_topic: netCDF-Java
 company_name: Unidata
 # this appears in the footer
 
-<<<<<<< HEAD
-=======
 github_editme_path:
-
->>>>>>> 8d42b06b4a... Version menu floats to the left!
-#github_editme_path: tomjohnson1492/documentation-theme-jekyll/blob/gh-pages/pages/
-# if you're using Github, provide the basepath to the branch you've created for reviews, following the sample here. if not, leave this value blank.
 
 disqus_shortname:
 # if you're using disqus for comments, add the shortname here. if not, leave this value blank.
--- a/docs/src/public/userguide/_config.yml
+++ b/docs/src/public/userguide/_config.yml
@@ -5,7 +5,7 @@ output: web
 docset_version: 5.4
 # this will appear in an HTML meta tag, sidebar, and perhaps elsewhere
 
-topnav_title: netCDF-Java Documentation
+topnav_title: netCDF-Java
 # this appears on the top navigation bar next to the home button
 
 site_title: netCDF-Java Documentation
@@ -17,6 +17,11 @@ site_topic: netCDF-Java
 company_name: Unidata
 # this appears in the footer
 
+<<<<<<< HEAD
+=======
+github_editme_path:
+
+>>>>>>> 8d42b06b4a... Version menu floats to the left!
 #github_editme_path: tomjohnson1492/documentation-theme-jekyll/blob/gh-pages/pages/
 # if you're using Github, provide the basepath to the branch you've created for reviews, following the sample here. if not, leave this value blank.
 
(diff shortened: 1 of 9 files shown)


ddirks:
Hi @oxelson! Thanks for noticing that . . . I'm not even sure how that particular change came about . . . @lesserwhirls has tried valiantly to get me up to speed with the git/github stuff in the last week, including walking me through a rebase yesterday (eek), on top of my own flailing. In any case it should end up with a URL of the form
`https://docs.unidata.ucar.edu/netcdf-java/5.3/userguide/feed.xml`
and I think it does now.

oxelson:
> Hi @oxelson! Thanks for noticing that . . . I'm not even sure how that particular change came about . . . @lesserwhirls has tried valiantly to get me up to speed with the git/github stuff in the last week, including walking me through a rebase yesterday (eek), on top of my own flailing. In any case it should end up with a URL of the form
> `https://docs.unidata.ucar.edu/netcdf-java/5.3/userguide/feed.xml`
> and I think it does now.

W00t!

lesserwhirls:
Looks great Doug!

You have the chance to do a rebase again now, although this one should be pretty clean (no conflicts). Let me know if you'd like to do that together.

When serving the files locally via Jekyll, Jekyll sets site.url to `host`:`port` (as defined in `_config.yml`),
so we get `http://127.0.0.1:4005/` when running locally using `./gradlew serveUsersGuide`. For grins, I built and published the docs from this PR and pushed them up to nexus:

https://docs.unidata.ucar.edu/netcdf-java/5.4/userguide/

Up there, it has:

```
<link rel="alternate" type="application/rss+xml" title="" href="www.unidata.ucar.edu/feed.xml">
```

Is that what we want? Or do we want:

```
<link rel="alternate" type="application/rss+xml" title="" href="https://docs.unidata.ucar.edu/netcdf-java/5.4/userguide/feed.xml">
```

In that case, I think we need to change `_config.yml` `docs/src/public/userguide/_includes/head.html`:

```
<link rel="alternate" type="application/rss+xml" title="{{ site.title }}" href="feed.xml">
```

Since the structure of the generated docset, html file wise, is flat, the relative link to `feed.xml` will always be correct.

Now, bigger question...do we even want an RSS feed for the docs? I'm ok if we just scrap that all together (remove the entry from `head.html`).

Now another big question - what about the sitemap?

https://docs.unidata.ucar.edu/netcdf-java/5.4/userguide/sitemap.xml

Because sitemap.xml uses site.url, we get entries like:

```
<loc>www.unidata.ucar.edu/basic_ncml_tutorial.html</loc>
```

That's totally fixable, but should each version of the docs have it's own sitemap? Or, when we do a release, do we specifically publish sitemap.xml to the "root" of the docset, so like `https://docs.unidata.ucar.edu/netcdf-java/sitemap.xml`? If so, would we generate links using /version/ or using /current/? Then we could point google at that sitemap?

ddirks:
Thanks Sean! I **knew** that running this on the real server would look different than the local Jekyl testing&hellip;

1. RSS feed. I clearly didn't pay enough attention to understanding this series of tweaks, but in the end I agree most with your final point. What is the purpose of an RSS feed on a pile of static html? The Jekyl template we're using includes provisions for commenting (which I suppose could make RSS useful), but we're not using them. So I'd advocate for just removing that from the doc sets.

1. Sitemap. I missed this bit entirely. I see two things:

    - As you say, these entries are generated from site.url, which points to Unidata's home. For each versioned docset, it seems like site.url needs to be the top of that particular docset's tree, e.g. `docs.unidata.ucar.edu/netcdf-java/5.3/userguide`
(this would fix the rss problem too, if we wanted an rss feed)
    - We should have a "master" sitemap.xml at the root of the doc server, which points to all the sitemap.xml files for individual docsets (including the "current" docsets) and then point Google to the "master" sitemap. Search results would then show the individual versioned sets, but presumably the "current" set would bubble to the top over time. The individual sitemap.xml files can be generated by the Jekyl builds, but the master will have to be generated some other way.

    Does this make sense to you, @oxelson?

1. Version picker code. Broken! My javascript to build the version menus to preserve the page you're currently on when switching to a different version does not work as I intended, so the version picker doesn't actually work on the docs server. I'll go fix that.

So, I propose to:

1. Remove the RSS feed from head.html
1. Update _config.yml to point to the root of the docset on docs.unidata.ucar.edu. (This adds to the list of things to be hand-altered when a new release is created.)
1. Fix the version picker javascript
1. Rebase again (I'll attempt this on my own, but may yell for help)
1. Ask for another review.

Sound okay?

oxelson:
>    * We should have a "master" sitemap.xml at the root of the doc server, which points to all the sitemap.xml files for individual docsets (including the "current" docsets) and then point Google to the "master" sitemap. Search results would then show the individual versioned sets, but presumably the "current" set would bubble to the top over time. The individual sitemap.xml files can be generated by the Jekyl builds, but the master will have to be generated some other way.

@ddirks  agreeded


> 1. Remove the RSS feed from head.html

Yay!  

> 2. Update _config.yml to point to the root of the docset on docs.unidata.ucar.edu. (This adds to the list of things to be hand-altered when a new release is created.)

Can the yml file be updated/generated with the build?  The less hand-altered files, the better.

lesserwhirls:
Sorry just now checking back in...gnarly interval variable subsetting work...at any rate, I agree with all of those changes. I made a few changes for the sitemap that didn't require adding the version to the url we put in `_config.yml`. I'd be happy to add those changes to this PR. We'd just need to coordinate to make sure we don't stomp all over each other.

ddirks:
No worries . . . I'm happy for a way to build the sitemap without adding the version number in a second place in _config.yml. Does your way involve changing site.xml? I don't see anything in your lesserwhirls/netcdf-java repo, but maybe I don't know where to look.

lesserwhirls:
I don't have it anywhere at the moment, but I can post it and let you know. Will be in an hour or so.

ddirks:
Okay, all the changes discussed above are included now.

lesserwhirls:
Beautiful! 

https://docs.unidata.ucar.edu/netcdf-java/5.4/userguide/

sitemap and sitemap_current look good as well. 

I'll backport these changes to `5.0`, `5.1`, `5.2`, and `5.3`, and we should be good to go. Thank you @ddirks!

Issue #152, opened 2019-11-20
Possible clarification of GRIB  2D Datasets with regards to CF


lesserwhirls:
The 2D GRIB collection datasets currently do something that is not recommended by CF. It's not a requirement, but it is recommended.

As an easy to see example (not a TDS issue, as the code lives in netCDF-Java in the `grib` module, but easier to see through the TDS), if we look at the cdl representation of the [GFS 80km dataset on thredds-test](https://thredds-test.unidata.ucar.edu/thredds/catalog/grib/NCEP/GFS/CONUS_80km/catalog.html?dataset=grib/NCEP/GFS/CONUS_80km/TwoD) (cdl [shown here]( https://thredds-test.unidata.ucar.edu/thredds/cdmremote/grib/NCEP/GFS/CONUS_80km/TwoD?req=cdl) via cdmremote), we see things like the following:

~~~
netcdf grib/NCEP/GFS/CONUS_80km/TwoD {
  dimensions:
    reftime = 124;
    time2 = 41;

  variables:
    double reftime(reftime=124);
      :units = "Hour since 2019-10-20T00:00:00Z";

    double time2(reftime=124, time2=41);
      :units = "Hour since 2019-10-20T00:00:00Z";

  float Pressure_surface(reftime=124, time2=41, y=65, x=93);
      :units = "Pa";
      :coordinates = "reftime time2 y x ";
~~~

The main issue here, I think, is with the use of `time1` (and other multidimensional coordinate variables), specifically that the variable `time1` is two dimensional, and there exists a dimension with the same name. The CF spec says:

> We recommend that the name of a multidimensional coordinate variable should not match the name of any of its dimensions because that precludes supplying a coordinate variable for the dimension. This practice also avoids potential bugs in applications that determine coordinate variables by only checking for a name match between a dimension and a variable and not checking that the variable is one dimensional.

(see http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#coordinate-system).

Strictly speaking, what we do here *is* CF compliant, but it can cause confusion for clients that do not check to see if the variable of a matching variable/dimension name pair meet the requirement of being a coordinate variable (that is, the variable is 1D), and so not recommended.

One thing we could do to help clarify the situation is to simply rename the `time*` dimensions to something like `time*_dim`, or rename the `time*` variable to something like `valid_time*`.

[...]

PR #282, opened 2020-05-06
Ensure 2D time variables do not use the name of a dimension

lesserwhirls:
Addresses https://github.com/Unidata/netcdf-java/issues/152


lesserwhirls:
Note: this passes the full set of tests run on Jenkins. w00t.

PR #284, opened 2020-05-07
Use new visad-mcidas-slim-ucar-ns artifact

lesserwhirls:
The `edu.wisc.ssec.mcidas` package can now be accessed via `visad-mcidas-slim`:

https://github.com/lesserwhirls/visad

The package name `edu.wisc.ssec.mcidas` has been relocated to `ucar.mcidas` in this particular artifact. This artifact is up-to-date with the latest changes in the main visad repository.


julienchastang:
Nice work. Did you have to muck with visad internals to get to this point such as cleaning up extremely old, baked in versions of netcdf-java? I also saw there is still `edu.wisc.ssec:visad:2.0-20130124` so maybe I am wishfully over interpreting. 😉

lesserwhirls:
Greetings @julienchastang! Actually, with this PR, there will only be one IOSP where you absolutely need the full visad jar, and that's the `Vis5DIosp` (which is isolated in the `visad/vis5d` subproject of netCDF-Java).

I was able to carve out the `edu.wisc.ssec.mcidas` package from visad, mostly intact:

https://github.com/lesserwhirls/visad

move the build to gradle, and publish those artifacts to our nexus server (details are over on that repo). This PR changes the area file related IOSPs (in the `visad/mcidas` subproject) to use that. The artifacts produced by lesserwhirls/visad do not have any dependencies outside of core java, which is a pretty big win I think.

lesserwhirls:
Jenkins passes - merging

PR #285, opened 2020-05-08
Add GRIB2 Template 4.32

ColemanTom:
This is just adding in the PDS part of adding in GRIB2 Template 4.32. I have not tackled items discussed in https://github.com/Unidata/thredds/issues/846#issuecomment-365959527 as I don't really know how the internals of this codebase work to provide a variable name containing information from the satellite series/number and wavelength (especially that obtains that information from the bufr tables, or a centre chosen alternative). But, all information is available from the template, so people can use this to construct their own naming convention.

A test was added to show that reading in of data works fine. I'm happy to shrink the test data more if preferred (it's currently a 10 degree by 10 degree subset of the globe).

I am raising this in response to https://github.com/Unidata/netcdf-java/issues/22#issuecomment-573401264 (I provided the data attached to that comment) as I made this (or near enough to this) as a patch to help them out directly if possible. I thought I should also raise it here.

Note also, I fixed up a problem in Pds30 and Pds31 which had the series and number octets mixed up (series comes before number) as per [WMO specifications](https://wmoomm.sharepoint.com/sites/wmocpdb/eve_activityarea/Forms/AllItems.aspx?id=%2Fsites%2Fwmocpdb%2Feve%5Factivityarea%2FWMO%20Codes%2FWMO306%5FvI2%2FAllVERSIONS%2F20200506%2FWMO306%5FvI2%5FGRIB2%5FTemplate%5Fen%2Epdf&parent=%2Fsites%2Fwmocpdb%2Feve%5Factivityarea%2FWMO%20Codes%2FWMO306%5FvI2%2FAllVERSIONS%2F20200506&p=true&originalPath=aHR0cHM6Ly93bW9vbW0uc2hhcmVwb2ludC5jb20vOmI6L3Mvd21vY3BkYi9FVW1uTE5BTTlXZE1yMVM3R1JNbF9HOEJGcXAtQjFRaWUtay12TXdtckcyMkdRP3J0aW1lPTdGU0dMZVh5MTBn).


lesserwhirls:
Very nice, and great catch on the `series` / `number` swap! 

Would you happen to have an example of GRIB messages of the same product valid at the same time but for different wavelengths? Some of the questions in Unidata/thredds#846 are about how the CDM should differentiate between the message that only differ by wavelength - by creating a new dimension for wavelength and storing them in the same variable, or by changing the the variable name to include the wavelength. I'm not asking you do that work, so no worries there - what you have done here gets us to a point to even tackle the other questions, so thank you! But if you have some sample GRIB message like I describe above, that'd be super helpful! Worst case, I can break out a hexeditor and...well...try to make some, but I'd rather deal with messages from the wild, so to say :-)

ColemanTom:
The grib file in the test has two messages in it, for the same basetime and forecast time, but different for all other relevant SatelliteBand metadata.

So, I don't think the file quite matches what you are wanting as the satelliteSeries is different. If you want, I can change the metadata for at least satelliteSeries/Number to make them consistent across both messages. Or, probably better, add another message to the sample file which does this (so you get a sample file with two messages with the same satellite but different wavelength, and two messages with the same wavelength but a different satellite.

For purposes of testing and development I don't mind the metadata being a bit off, and this would only take me a minute to do.

ColemanTom:
Oh, out of curiousity, is there a standard way that synthetic satellite data is described in netCDF metadata? I know netCDF is less rigid than grib2, so I was wondering if you've seen examples of how people have approached it?

lesserwhirls:
> Or, probably better, add another message to the sample file which does this (so you get a sample file with two messages with the same satellite but different wavelength, and two messages with the same wavelength but a different satellite.

That would be super helpful - thank you!

> Oh, out of curiousity, is there a standard way that synthetic satellite data is described in netCDF metadata?

That's a great question, but not one that I know the answer to myself. @ethanrd - do you know of anything in particular related to synthetic satellite data?

At any rate, thank you again for this contribution. I'll go ahead and merge this is, and will see what needs to be one in terms of handling multiple wavelengths.

ethanrd:
Hi @lesserwhirls & @ColemanTom - I'm afraid I don't know of any netCDF standards specifically for synthetic satellite data.

cofinoa:
@ColemanTom you can take a look at this draft of proposal for encoding satellite swath data:

https://github.com/Unidata/EC-netCDF-CF/blob/master/swath/swath.adoc

it's not for synthetic data, but it could be a start point.

ColemanTom:
@cofinoa thanks I'll have a read.

PR #289, opened 2020-05-13
deprecate FileCacheable

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/289)

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/NetcdfFile.java
+++ b/cdm/core/src/main/java/ucar/nc2/NetcdfFile.java
@@ -880,6 +880,7 @@ public class NetcdfFile implements FileCacheable, Closeable {
    * 
    * @deprecated do not use
    */
+  @Deprecated
   public void release() throws IOException {
     if (iosp != null)
       iosp.release();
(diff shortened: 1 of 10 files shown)


lesserwhirls (commented):
Have you ever heard of anyone implementing their own FileCache (edit: specifically for use by netCDF-Java)?

> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/289#-:-M7IdHFw1Jfw8Gp77x46:b-ctclcr)* status: :shipit: complete! all files reviewed, all discussions resolved

JohnLCaron:
No, though its possible. 
The idea is that caching should be independent of the cached objects.
Much of the design was predicated on the idea that it is expensive to open files. Later testing showed that (some?) OS are caching open file handles. So this whole thing may be a case of premature optimization. In any case, it deserves further timing tests, and should not be in the API, but an implementation decision that can change without the user caring.

lesserwhirls:
Sounds good. My guess is that at this point, between improvements to Java, OS enhancements, and bigger/better cache on drives, we're getting little for a lot of and mental address space, so to say (late 90s, early-to-mid 2000s was a different story, I'm sure).

lesserwhirls:
Thanks @jlcaron!

PR #301, opened 2020-05-28
Cross-seam coverage subset with a horizontal stride

lesserwhirls:
The coverage subset's longitude axis was not taking into account the presence of a horizontal stride when it was constructed. As the saying goes

> index space considered harmful

and the idea of setting a horizontal stride feels a little too close to index space.


lesserwhirls:
Failures due to jcenter.bintray.com 504. Re-running.

lesserwhirls:
jcenter is still have issues. Will try again after bit.

lesserwhirls:
...w00t?

PR #305, opened 2020-05-29
Replace ProjectionPointImpl with ProjectionPoint.create.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/305)


rschmunk:
Oooh, this might eliminate 95% of the deprecation notices I've been getting when linking with recent NJ versions.

lesserwhirls:
Looks good to me - thanks @JohnLCaron!

rschmunk:
I spoke too soon. This is great, but it's the deprecation of `ProjectionImpl` that is throwing 28 warnings out of currently 54 when I compile/link my code against a snapshot from 6/1. So I guess I'll have to wait a bit more.

JohnLCaron:
Hi Robert. Yes, I cant change  ProjectionImpl to  Projection everywhere
until ver6, since I cant break the API yet. However, you can change any
ProjectionImpl  return types to  Projection now, which will prepare for
ver6. John

rschmunk:
@JohnLCaron, Per e-mail I sent to @lesserwhirls earlier, I belatedly realized that the way my code was working, I didn't really need `ProjectionImpl`.

Since I was creating objects by calling `CoordinateSystem.getProjection`, I could just declare them as type `Projection` instead. D'oh!

So at the moment, I have managed to reduce the number of deprecation warnings I get when compiling against NJ down to 22.

PR #317, opened 2020-06-06
Fix file being unlocked in cache before resources fully released

### LINKED ISSUE #316: NullPointerException when reading cached files concurrently
Tested in versions: 5.4.0-SNAPSHOT, 5.3.2, 5.2.0
Environment: OpenJDK 1.8, CentOS 7.x

When reading concurrently from files returned by cache, a `NullPointerException` often occurs:
```
Exception in thread "Thread-20" java.lang.NullPointerException
	at ucar.nc2.iosp.IospHelper.readData(IospHelper.java:96)
	at ucar.nc2.iosp.IospHelper.readDataFill(IospHelper.java:51)
	at ucar.nc2.internal.iosp.netcdf3.N3iospNew.readData(N3iospNew.java:322)
	at ucar.nc2.internal.iosp.netcdf3.N3iospNew.readData(N3iospNew.java:178)
	at ucar.nc2.NetcdfFile.readData(NetcdfFile.java:2121)
	at ucar.nc2.Variable.reallyRead(Variable.java:797)
	at ucar.nc2.Variable._read(Variable.java:736)
	at ucar.nc2.Variable.read(Variable.java:614)
	at CacheTest.readData(CacheTest.java:53)
	at CacheTest.run(CacheTest.java:40)
	at java.lang.Thread.run(Thread.java:745)
```
It seems like a race condition where two threads are able to acquire the same `NetcdfFile`.

PoC program using [this](https://github.com/Unidata/netcdf-java/blob/master/cdm/core/src/test/data/ncml/nc/cldc.mean.nc) test data:
```java
import java.io.IOException;

import ucar.nc2.NetcdfFile;
import ucar.nc2.Variable;
import ucar.nc2.dataset.DatasetUrl;
import ucar.nc2.dataset.NetcdfDatasets;

public class CacheTest implements Runnable {
	
	private String path = "res/cldc.mean.nc";
	
	public static void main(String[] args) throws Exception {
		NetcdfDatasets.initNetcdfFileCache(10, 25, 60);
		
		Runtime.getRuntime().addShutdownHook(new Thread() {
			@Override
			public void run() {
				NetcdfDatasets.shutdown();
			}
		});
		
		int numThreads = 25;
		Thread[] threads = new Thread[numThreads];
		for (int threadIdx = 0; threadIdx < numThreads; threadIdx++) {
			threads[threadIdx] = new Thread(new CacheTest());
			threads[threadIdx].start();
		}
		for (int threadIdx = 0; threadIdx < numThreads; threadIdx++) {
			threads[threadIdx].join();
		}
	}

	@Override
	public void run() {		
		for (int i = 0; i < 500; i++) {
			NetcdfFile file = null;
			try {
				DatasetUrl location = DatasetUrl.findDatasetUrl(path);
				file = NetcdfDatasets.acquireFile(location, null);
				readData(file);
			} catch (IOException e) { 
				e.printStackTrace();
			} finally {
				if (file != null) { 
					try { file.close(); } catch (Exception e) { e.printStackTrace(); }
				}
			}
		}
	}
	
	private void readData(NetcdfFile file) throws IOException {
		Variable var = file.findVariable("cldc");
		var.read();
	}
}
```

JMdoubleU:
This _might_ resolve #316. The NPE occurs very inconsistently but I haven't observed it with this change.

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/util/cache/FileCache.java
+++ b/cdm/core/src/main/java/ucar/nc2/util/cache/FileCache.java
@@ -466,8 +466,15 @@ public class FileCache implements FileCacheIF {
       }
       file.lastAccessed = System.currentTimeMillis();
       file.countAccessed++;
-      file.ncfile.release();
-      file.isLocked.set(false);
+
+      try {
+        file.ncfile.release();
+        file.isLocked.set(false);
+      } catch (IOException ioe) {
+        cacheLog.error("FileCache {} release failed on {} - will try to remove from cache. Failure due to:", name,
+            file.getCacheName(), ioe);
+        remove(file);
+      }
 
       if (cacheLog.isDebugEnabled())
         cacheLog.debug("FileCache " + name + " release " + ncfile.getLocation() + "; hash= " + ncfile.hashCode());
--- a/cdm/core/src/main/java/ucar/nc2/util/cache/FileCache.java
+++ b/cdm/core/src/main/java/ucar/nc2/util/cache/FileCache.java
@@ -466,8 +466,8 @@ public class FileCache implements FileCacheIF {
       }
       file.lastAccessed = System.currentTimeMillis();
       file.countAccessed++;
-      file.isLocked.set(false);
       file.ncfile.release();
+      file.isLocked.set(false);
 
       if (cacheLog.isDebugEnabled())
         cacheLog.debug("FileCache " + name + " release " + ncfile.getLocation() + "; hash= " + ncfile.hashCode());


lesserwhirls:
Nice catch! For sure, the lock should not be removed until the very end.

It's possible that `file.ncfile.release()` will throw an `IOException`, in which case we'd end up with an entry in the cache that will never be unlocked, and depending on how the upstream application handles `IOException`s, that could be bad over time. I wonder if it'd be worth catching the first `IOException` and chaining it to a new one that indicates the problem?

JMdoubleU:
@lesserwhirls In that case perhaps the caller would want to treat the file as being in a bad state and remove it from the cache, but it seems there's currently no public methods that can accomplish that for a specific file? Maybe it'd make sense to just remove it as part of the `release(...)` method if an exception is thrown?

lesserwhirls:
Good point. The cache has an eject method that will try to remove an entry from the cache. Since that's about the only thing you could do in this situation as the caller, perhaps it makes sense to have let the cache try to manage the exception itself. Something like:

~~~java
try {
  file.ncfile.release();
  file.isLocked.set(false);
} catch (IOException ioe) {
  log.error("FileCache " + name + " release failed on " + file " - will try to remove from cache.", ioe);
  remove(file);
}
~~~

If the remove fails, that will log an error as well.

lesserwhirls:
I went ahead and made the adjust, and it will be in the next release (`v5.3.3`, hopefully later today). Thank you for your contribution @JMdoubleU!

PR #321, opened 2020-06-09
Fix CompositeStationCollection dangling file locks

### LINKED ISSUE #320: CompositeStationCollection leaves files locked
There are cases where the code in `CompositeStationCollection` will leave a `FileCachable` locked as it searches through a multifile collection for Station Features, specifically when that feature does not exist in a given file. This was first noticed on the TDS side (see Unidata/tds#88).

lesserwhirls:
Fixes Unidata/netcdf-java#320. Special thanks to @ktyle for grinding away to come up with a set of files and an NCSS request that triggered the bug from the TDS side!


ktyle:
Thanks so much @lesserwhirls !! This is a really nice fix!!

PR #324, opened 2020-06-12
Array as Value Object

hlfsousa:
- Array and Index implement java.io.Serializable
- Tests for serialization of arrays of primitive and string
- Small correction to failure when Array.makeFromJavaArray with a String array

The purpose of this set of changes is to allow Array to be used as serializable Value Objects. Some systems and frameworks (e.g. Apache Storm/Kafka) require that objects be serializable. If Array has such capability it can be used directly when assembling NetCDF files in parts. There is otherwise no functional change in the classes involved, and no performance impact.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/324)

### DIFF
--- a/cdm/core/src/main/java/ucar/ma2/Array.java
+++ b/cdm/core/src/main/java/ucar/ma2/Array.java
@@ -5,6 +5,7 @@
 package ucar.ma2;
 
 import javax.annotation.Nonnull;
+import java.io.Serializable;
 import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


JohnLCaron:
Hi hlfsousa:

Following Josh Bloch's advice (eg Effective Java Chapter 12 Serialization) and articles such as https://dzone.com/articles/jdk-11-beginning-of-the-end-for-java-serialization, we are planning on removing object serialization (possible in the transition of Netcdf version 6 to version 7), and likely replacing it with google protobuf serialization. The big problem with serialization is there are lots of security problems.

It would be good to hear about the needs of Apache Storm/Kafka, and if protobuf is something that would work in that context.

In the meanwhile, we could still consider allowing this PR for version 5 and 6, given that the other versions are a year or so away. But with the understanding that it will all go away in version 7. Version 6 would likely contain both object serialization and protobuf serialization to aid in the transition. 

Also see https://github.com/lesserwhirls/netcdf-java-api-testing/wiki/Roadmap-for-Netcdf-Java

John Caron

PR #333, opened 2020-06-18
Fix geostationary scaling

rschmunk:
If grid x/y units were not radians but something else compatible (e.g., microradians), appropriate scaling of the transform in `ucar.unidata.geoloc.projection.sat.Geostationary` was not done.

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/dataset/transform/Geostationary.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/transform/Geostationary.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998-2018 University Corporation for Atmospheric Research/Unidata
+ * Copyright (c) 1998-2020 University Corporation for Atmospheric Research/Unidata
  * See LICENSE for license information.
  */
 
@@ -66,29 +66,45 @@ import ucar.unidata.geoloc.ProjectionImpl;
  * @since 12/5/13
  */
 public class Geostationary extends AbstractTransformBuilder implements HorizTransformBuilderIF {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Geostationary.class);
 
   private static double defaultScaleFactor = -1.0;
 
+  /**
+   *
+   */
   public String getTransformName() {
     return CF.GEOSTATIONARY;
   }
 
+  /**
+   *
+   */
   public TransformType getTransformType() {
     return TransformType.Projection;
   }
 
+  /**
+   *
+   */
   private double getScaleFactor(String geoCoordinateUnits) {
     // default value of -1.0 interpreted as no scaling in the class
     // ucar.unidata.geoloc.projection.sat.Geostationary
     double scaleFactor = defaultScaleFactor;
     String neededMapCoordinateUnit = "radian";
+
     if (SimpleUnit.isCompatible(geoCoordinateUnits, neededMapCoordinateUnit)) {
       scaleFactor = SimpleUnit.getConversionFactor(geoCoordinateUnits, neededMapCoordinateUnit);
     }
 
+    logger.debug("geoCoordinateUnits {}, scaleFactor {}", geoCoordinateUnits, scaleFactor);
+
     return scaleFactor;
   }
 
+  /**
+   *
+   */
   public ProjectionCT makeCoordinateTransform(AttributeContainer ctv, String geoCoordinateUnits) {
     readStandardParams(ctv, geoCoordinateUnits);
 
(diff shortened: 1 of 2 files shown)


lesserwhirls:
Thank you for taking a look at this. Now I need to figure out why things looked ok in ToolsUI. If you'd like, I can fix the spotless failure so you won't have to mess with that. I'd also remove the empty comment blocks, and fix the javadoc for latLonToProj (currently it says it returns values with units in radians).

rschmunk:
I just ran spotless.

Personally, I find the empty comment blocks makes things more readable, as it helps show where method definitions begin when you're scrolling down the code.

rschmunk:
Sorry, did not notice the comment about units in latLonToProj.

Changing the returned units is what got my plotting to work. Perhaps ToolsUI expects what it returns to be in radians?

rschmunk:
@lesserwhirls, Is there a way to make a color 2D plot of gridded data in ToolsUI? I thought I'd been able to do so in the past, but in trying to doublecheck its handling of geostationary gridded dataset I couldn't figure out how, assuming it is possible..

lesserwhirls:
Greetings @rschmunk!

ToolsUI can show a simple 2D projected image through the `FeatureTypes` &rarr; `Grids` tab. Once the dataset is open, you have to click the Red Alien icon:

ft-061912

and that will open the grid viewer. Once you select a grid, you can click the alien again and it will display:

gv-061912

I made that with your PR, and it looks great. With the PR I made earlier, it looked ok at first glance, but the values were a few degrees off. I have an idea for a test to make sure this doesn't accidentally break again, but I can do that in another PR.  I also want to make sure we're testing the `crossSeam` method as well.

I can understand the readability aspect with the empty blocks, and I've heard arguments either way (not quite at the level of tabs vs spaces, though!). My concern with the empty blocks is that since it comes down to personal preference (again, nothing wrong with that), reviewing the diffs on a PR can quickly become overwhelmed with changes related to adding/removing them, and it makes the process of reviewing a PR a little more difficult in terms of keeping track of what actual changes were made to fix a bug (especially for large changesets). We went with the convention of not having them, but I understand they are useful to you and your development process. Are you generating those by hand or using your IDE? If they take very little on your part to generate, I could tweak spotless to look for them and remove them once spotlessApply is run? With that workflow, a contributor would still get the benefit of having them as they develop locally if that was their preference, but the diffs on github would focus in on the code changes after applying spotless.

rschmunk:
@lesserwhirls, Somehow in seeing the tabbed named FeatureType, I thought it would be related to trajectories and other discrete types. But nevertheless, all I could get that panel to display was a colorbar alongside a blank plot.

Anyway, I updated the PR so that the javadoc for lonlatToProj describes the units of the returned x/y values, and also to remove all the blank comment blanks.

lesserwhirls:
I think there a few cases when work done in netCDF-Java served as a proof-of-concept for proposed extensions to CF, and that language, I think, is an example of that. What ended up as the discrete sampling geometries of CF started life as point FeatureTypes in netCDF-Java:

https://cf-pcmdi.llnl.gov/trac/ticket/37

The `FeatureType` tab in toolsUI refers to "Scientific Feature Type", which is the highest layer in the Common Data Model:

https://docs.unidata.ucar.edu/netcdf-java/current/userguide/common_data_model_overview.html

The `Viewer` tab roughly corresponds to the top of the Data Access Layer (`NetcdfFile`), the `CoordSystem` tab to the top of the Coordinate System Layer (i.e. `NetcdfDataset`), and the `FeatureTypes` tab to the various types in the Scientific Feature Types layer (e.g. `Coverage`, `grid`, `fmrc`, `radial`, `point`, etc.).

It all can be a bit confusing.

At any rate, *thank you* again for this fix! I know that code isn't easy to follow. If I had all the time in the world, I'd love to extend `ProjectionPoint`, `LatLonPoint`, and perhaps add new flavor of `Array` to have the option of holding unit information in the Coordinate System Layer. That could allow for the various conventions to attach units to global attribute values, for example, and then let coordinate system builders use that information without needing the various one-offs for passing around unit strings. Same kind of thing when passing in points to `latLonToProj` and `projToLatLon` - sure would be nice to know the unit attached to those.

rschmunk:
Thanks @lesserwhirls. As this PR changed functionality of the `latLonToProj` method a bit, I wasn't sure whether you'd be in much of a hurry to accept it.

The original "problem" such as it was could have instead been blamed on me not reading the original javadoc. On the other hand, I personally think it's important for `latLonToProj` and `projToLatLon` to use the same units for the X/Y `ProjectionPoint` object.

That's an interesting thought that `ProjectionPoint` might be extended to also include info about the units of the X/Y values.

lesserwhirls:
Part of my confusion surrounding this is that there isn't a requirement that every projection coordinate variable that uses a particular transform must use the same units. For these GOES products, we only have one set of x and y coordinate variables, so we're ok to use the unit we find. I agree, though, that it makes more sense to make `latLonToProj` produce a number that can be fed into `projToLatLon` when using the same projection object in the general case.

PR #339, opened 2020-06-30
Fix jenkins test

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/339)

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/stream/TestStreamWriter.java
+++ b/cdm-test/src/test/java/ucar/nc2/stream/TestStreamWriter.java
@@ -61,7 +61,8 @@ public class TestStreamWriter {
   public void testN3outputStreamWriter() throws IOException {
     System.out.println("\nFile= " + endpoint + " size=" + new File(endpoint).length());
     long start = System.currentTimeMillis();
-    NetcdfFile fileIn = NetcdfFiles.open(endpoint);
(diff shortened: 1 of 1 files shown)


JohnLCaron:
Failure probably unrelated to PR.

lesserwhirls:
Yeah, the github actions failure isn't related.

It looks like this test fails when using `NetcdfFiles` because we are not telling the iosp to make the record structure variable, but in `NetcdfFile` [we do](https://github.com/Unidata/netcdf-java/blob/963e41e1bf56e7a09afa0a2c11edc946c9d82bb1/cdm/core/src/main/java/ucar/nc2/NetcdfFile.java#L1930-L1932).

The NPE is because the variable `record` does not exist (because it's not created), and `N3outputStreamWriter` [expects one to exist](https://github.com/Unidata/netcdf-java/blob/963e41e1bf56e7a09afa0a2c11edc946c9d82bb1/cdm/core/src/main/java/ucar/nc2/iosp/netcdf3/N3outputStreamWriter.java#L59). Of course, `N3outputStreamWriter` is marked as experimental and deprecated, so maybe it's all good.

JohnLCaron:
Yes, I decided not to actually fix the test, but just reverted. So without the github action failure, this should be ready to go, AFAIK.

Ill have to review status of N3outputStreamWriter, but I think it was never used and will be removed.

PR #350, opened 2020-07-07
Clarify jni netcdf4 public API.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/350)

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/ft/coverage/TestCoverageFileWriterP.java
+++ b/cdm-test/src/test/java/ucar/nc2/ft/coverage/TestCoverageFileWriterP.java
@@ -18,7 +18,7 @@ import ucar.nc2.dataset.NetcdfDatasets;
 import ucar.nc2.dt.grid.GridDataset;
 import ucar.nc2.ft2.coverage.*;
 import ucar.nc2.ft2.coverage.writer.CFGridCoverageWriter;
-import ucar.nc2.jni.netcdf.Nc4Iosp;
+import ucar.nc2.jni.NetcdfClibrary;
 import ucar.nc2.write.NetcdfFileFormat;
 import ucar.nc2.write.NetcdfFormatWriter;
 import ucar.unidata.util.test.category.NeedsCdmUnitTest;
@@ -85,7 +85,7 @@ public class TestCoverageFileWriterP {
   @Test
   public void writeTestFile() throws IOException, InvalidRangeException {
     // skip test requiring netcdf4 if not present.
-    if (version.isNetdf4format() && !Nc4Iosp.isClibraryPresent()) {
+    if (version.isNetdf4format() && !NetcdfClibrary.isClibraryPresent()) {
       return;
     }
     System.out.printf("Test Dataset %s type %s%n", endpoint, type);
--- a/cdm-test/src/test/java/ucar/nc2/ft/coverage/TestCoverageFileWriterP.java
+++ b/cdm-test/src/test/java/ucar/nc2/ft/coverage/TestCoverageFileWriterP.java
@@ -18,7 +18,7 @@ import ucar.nc2.dataset.NetcdfDatasets;
 import ucar.nc2.dt.grid.GridDataset;
 import ucar.nc2.ft2.coverage.*;
 import ucar.nc2.ft2.coverage.writer.CFGridCoverageWriter;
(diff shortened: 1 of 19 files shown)


lesserwhirls (changes requested):
A few comments to consider. Thanks @johnlcaron!

> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/350#-:-MBd8rDJEwlbfkOz-2KJ:bgjadzq)* status: 0 of 17 files reviewed, 4 unresolved discussions (waiting on @JohnLCaron)

___
*[netcdf4/src/main/java/ucar/nc2/jni/NetcdfClibrary.java, line 73 at r1](https://reviewable.io/reviews/unidata/netcdf-java/350#-MBd3ylt-_NZfvKf7Z9U:-MBd3ylu9U9cowjLOSDu:bvnbl8z) ([raw file](https://github.com/unidata/netcdf-java/blob/e1328be75110af0112450d37f11d50fdec264593/netcdf4/src/main/java/ucar/nc2/jni/NetcdfClibrary.java#L73)):*
> ```Java
>    * @return true if present
>    */
>   public static synchronized boolean isClibraryPresent() {
> ```

I can't quite put my finger on it (maybe the repetition of it following `NetcdfClibrary`), but what about `isLibraryPresent()`, or even `isPresent()`, instead of `isClibraryPresent()`?
___
*[netcdf4/src/main/java/ucar/nc2/jni/NetcdfClibrary.java, line 81 at r1](https://reviewable.io/reviews/unidata/netcdf-java/350#-MBd54lUC0e44m8pev1L:-MBd54lUC0e44m8pev1M:bnp092s) ([raw file](https://github.com/unidata/netcdf-java/blob/e1328be75110af0112450d37f11d50fdec264593/netcdf4/src/main/java/ucar/nc2/jni/NetcdfClibrary.java#L81)):*
> ```Java
> 
>   /** Get the interface to the Nectdf C library. */
>   public static synchronized Nc4prototypes getClibrary() {
> ```

Same as `isClibraryPresent()` - maybe `getLibrary()`, or `getNativeMethods()`?
___
*[netcdf4/src/main/java/ucar/nc2/jni/package.html, line 7 at r1](https://reviewable.io/reviews/unidata/netcdf-java/350#-MBd6zaP3hBw_Uazy6GB:-MBd6zaP3hBw_Uazy6GC:brq3l7n) ([raw file](https://github.com/unidata/netcdf-java/blob/e1328be75110af0112450d37f11d50fdec264593/netcdf4/src/main/java/ucar/nc2/jni/package.html#L7)):*
> ```HTML
> <HTML>
> <HEAD>
>   <TITLE>package ucar.nc2.jni</TITLE>
> ```

Rather than having a particular interface name in the package (`jni`, `jna`), maybe we just move it in 6 to use the more generic term `native`. Given the number of HDF5 filters without java versions, I have a feeling we'll need to start wrapping more native code at some point in the near future, so I could even see moving to `ucar.nc2.native.netcdf` and `ucar.nc2.native.filters` at some point.

On a related note - it might be worth renaming the artifact to something like `netcdf-native` (in 6). I see a lot of people pulling in the netcdf4 artifact when they only need to read.
___
*[netcdf4/src/main/java/ucar/nc2/jni/netcdf/SizeT.java, line 15 at r1](https://reviewable.io/reviews/unidata/netcdf-java/350#-MBd224E0uIhnkxM7eh0:-MBd2-4Q4CFZjgO1O_76:bei12os) ([raw file](https://github.com/unidata/netcdf-java/blob/e1328be75110af0112450d37f11d50fdec264593/netcdf4/src/main/java/ucar/nc2/jni/netcdf/SizeT.java#L15)):*
> ```Java
>  *
>  * @see "https://github.com/twall/jna/issues/191"
>  *      TODO: This will move packages in version 6.
> ```

Looks like the jna issue has since been addressed recently?

https://github.com/java-native-access/jna/commit/d471b15353a9dbcc5a29064385e4e50ca5d257c5

Might be able to get rid of SizeT and SizeTByReference at some point soon. Worth adding a note about that?

JohnLCaron:
These are all good suggestions. However, native is reserved and cant be used. Its also a bit overloaded.
Trying to find some other word. how about ffi = foreign function interface (see jni wikipedia article)? So use that instead of jni?

JohnLCaron:
ready if you are

PR #357, opened 2020-07-13
Last updates on master

lesserwhirls:
Last updates on master - after this merges in, will make a new branch called `develop` as version `6.0.0-SNAPSHOT`, which will become the repo default branch, and `master` will become `maint-5.x`.


lesserwhirls:
I'll deal with the TDS failure in a bit - it's related to metadata parsing in the new pom files, but I think it's a bug in gradle 3.5...only one thing to do there, and that's upgrade the tds build (next on the list).

PR #367, opened 2020-07-17
Add Java 11 to the test matrix

lesserwhirls:
Even though we require Java 8, since we've upgraded Gradle, we can now start testing on Java 11. Let's see how this goes.

### DIFF
--- a/.github/workflows/netcdf-java.yml
+++ b/.github/workflows/netcdf-java.yml
@@ -4,12 +4,16 @@ jobs:
   tests-zulu:
     name: netCDF-Java Tests (Zulu JDK)
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # test against latest 8 and 11
+        java: [ 8, 11 ]
     steps:
       - uses: actions/checkout@v2
-      - name: Setup Zulu JDK 1.8
+      - name: Setup java ${{ matrix.java }}
         uses: actions/setup-java@v1
         with:
-          java-version: 1.8
+          java-version: ${{ matrix.java }}
           architecture: x64
       - name: Install netCDF-C
         run: sudo apt update && sudo apt-get install libnetcdf-dev
@@ -23,10 +27,10 @@ jobs:
         run: ./gradlew --info --stacktrace testAll
         env:
           TRAVIS: 'true'
-      - uses: actions/upload-artifact@v1
+      - uses: actions/upload-artifact@v2
         if: failure()
         with:
-          name: netCDF-Java_JUnit_Results_${{ github.sha }}_Zulu-JDK-8
+          name: netCDF-Java_JUnit_Results_${{ github.sha }}_Zulu-JDK-${{ matrix.java }}
           path: build/reports/allTests
 
   tests-adpotopenjdk-hs-8:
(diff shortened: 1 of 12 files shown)


lesserwhirls:
Hmmmm....now this is interesting.

Let's look at one failure (`ucar.unidata.geoloc.TestProjections#testAEA`):

~~~java
AlbersEqualArea p = new AlbersEqualArea();
AlbersEqualArea p2 = (AlbersEqualArea) p.constructCopy();
assertThat(p).isEqualTo(p2);
~~~

The failure comes in when comparing the `lon0` values stored in `p` and `p2`.  `lon0` is actually a calculated value:

~~~java
public AlbersEqualArea(double lat0, double lon0, double par1, double par2, double falseEasting, double falseNorthing,
      double earth_radius) {
    super("AlbersEqualArea", false);

    this.lon0Degrees = lon0;
    this.lat0 = Math.toRadians(lat0);
    ...
~~~

and in the constructCopy() method, we end up calling `getOriginLon()`, like this:

~~~java
public Projection constructCopy() {
    return new AlbersEqualArea(getOriginLat(), getOriginLon(), getParallelOne(), getParallelTwo(), getFalseEasting(),
        getFalseNorthing(), getEarthRadius());
  }
~~~

and `getOriginLon()` does this:

~~~java
  public double getOriginLon() {
    return Math.toDegrees(lon0);
  }
~~~

So essentially, the we end up doing the following in our `.equals` method:

~~~
Math.toDegrees(Math.toRadians(lon0)) == lon0
~~~

That's...well, I wish that worked.

Just for fun, you can take this code:

~~~java
import java.util.stream.IntStream;
import java.util.List;
import java.util.stream.Collectors;

class RadDegBug {
  public static void main(String[] args) {
    IntStream stream = IntStream.range(-30, 30);
    List<Integer> wtf = stream.parallel().filter(i -> Math.toDegrees(Math.toRadians(i)) != i).mapToObj(Integer::valueOf)
        .sorted().collect(Collectors.toList());
    System.out.println(wtf);
  }
}
~~~

and compile and run it using JDK 8 and JDK 11. The output changes:

JDK 8 (OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.242-b08, mixed mode))
~~~posh
[-30, -26, -22, -15, -13, -11, 11, 13, 15, 22, 26]
~~~

JDK 11 (OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.7+10, mixed mode))
~~~posh
[-30, -29, -24, -15, -12, -6, -3, 3, 6, 12, 15, 24, 29]
~~~

So looks like we were lucky in JDK 8, but not so much JDK 11. Based on the comments I see in `ucar.unidata.geoloc.TestProjections` (e.g. `// LOOK this fails`), my guess is this has bit us before in previous JDK upgrades. 

Now, if I modify the code to use the actual values put into the constructor in the `equals` and `hashCode` methods, we're golden. Maybe this will fix the other two failures as well? We will see...but most likely tomorrow.

lesserwhirls:
The Zulu JDK tests did pass - looks like github is waiting for the old Zulu check. I'll update the required statuses after this.

PR #370, opened 2020-07-20
Java 11 support for 5.x branch

lesserwhirls:
This PR is a back port of the fixes in https://github.com/Unidata/netcdf-java/pull/367.


lesserwhirls:
There were a few incompatibilities between groovy and  java 11 on this branch that we didn't see on `develop` - basically, groovy + java 11 did not know what to do with the static method `EarthLocation` in the interface `EarthLocation`. Bumping groovy fixed that (went ahead and bumped spock as well). Will make the same changes on `develop` next, as maintenance.

PR #390, opened 2020-07-27
Remove / cleanup attribute deprecations.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/390)


JohnLCaron:
ready

JohnLCaron:
Note that fmrc and opendap are likely now broken. Im leaving them in the compile for now in order to make as many fixes as possible. At some point (likely when we change to true Immutable) all the unported modules likely have to be disabled.

lesserwhirls:
I believe these commits are reflected in Unidata/netcdf-java#391.

PR #426, opened 2020-08-11
Cleanup nc2.iosp, add tests.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/426)


JohnLCaron:
ready

lesserwhirls (approved):
> Reviewed 51 of 51 files at r1, 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/426)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnLCaron)

___
*[cdm-test-utils/src/main/java/ucar/unidata/util/test/TestDir.java, line 85 at r1](https://reviewable.io/reviews/unidata/netcdf-java/426#-MEXZm3tDQzGz7sZbN4E:-MEXZm3uBMRYLeA3PWND:bc2uz4v) ([raw file](https://github.com/unidata/netcdf-java/blob/1f1d6193fcfda9d10b9a7a847ff0ac115fc6f061/cdm-test-utils/src/main/java/ucar/unidata/util/test/TestDir.java#L85)):*
> ```Java
>   public static String cdmLocalTestDataDir = "../core/src/test/data/";
> 
>   /** The cdm-core local test data, use from cdm-test submodule. */
> ```

Oh man, I was just looking around in here at the various test directory fields. We could let gradle set a system property that would give us the absolute path to the repository root, which we could then use to set a subset of these without the need for relative paths. Not something you'd need to do in this PR...just tossing out the idea.

JohnLCaron:
Yeah, theres a lot of code touches to implement that, but we should do it when theres time.

Issue #413, opened 2020-08-05
Preload remote data results in HTTP 414 error response


rschmunk:
Working with a recent NJ 5.4 snapshot.

A user testing a data catalog server reported that using Panoply, he was getting an HTTP 414 error response. In checking into this, I also encounter the issue with toolsUI. I am unable to test using IDV as that app is not notarized to work on up-to-date Macs, but I assume the same would happen there.

The test dataset is presently [located here](http://balto.opendap.org/opendap/nsidc/ATL08_20181016124656_02730110_002_01.h5).

The exception that occurs is in `DODSNetcdfFile`, where an attempt to preload data creates a URL that is exceptionally long. Exceptionally long, as in 13kB, and so the GET request is rejected.

A pertinent logging excerpt from Panoply follows, omitting the full GET query string:

```
000976 SEVERE: DODSNetcdfFile preloadData - ERROR readDataDDSfromServer on ?[13617 chars omitted]
000977 WARNING: NcDataset acquireNjDataset - IOException acquiring http://balto.opendap.org/opendap/nsidc/ATL08_20181016124656_02730110_002_01.h5, enhance true
java.io.IOException: Method failed:HTTP/1.1 414 Request-URI Too Long on URL= http://balto.opendap.org/opendap/nsidc/ATL08_20181016124656_02730110_002_01.h5.dods?[13617 chars omitted]
	at ucar.nc2.dods.DODSNetcdfFile.preloadData(DODSNetcdfFile.java:1538)
	at ucar.nc2.dods.DODSNetcdfFile.<init>(DODSNetcdfFile.java:328)
	at ucar.nc2.dods.DODSNetcdfFileProvider.open(DODSNetcdfFileProvider.java:24)
	at ucar.nc2.dataset.NetcdfDatasets.openProtocolOrFile(NetcdfDatasets.java:431)
	at ucar.nc2.dataset.NetcdfDatasets.openDataset(NetcdfDatasets.java:152)
	at ucar.nc2.dataset.NetcdfDatasets.acquireDataset(NetcdfDatasets.java:280)
	at ucar.nc2.dataset.NetcdfDatasets.acquireDataset(NetcdfDatasets.java:257)
	...
```

So is this something where `DODSNetcdfFile` should be smart enough to reduce the size of the query strings it is issuing? Or should it even be requesting that much when the remote file is first accessed? Or a problem on the catalog server config? Or... any suggestions?

lesserwhirls:
[...]

PR #430, opened 2020-08-13
Manage size of dap GET request prefetch URI

lesserwhirls:
When prefetching data for a dap request, we need to manage the size of
the URI to avoid a "414 URI Too Long" response from the server. Tricky
because the maximum size of a GET request URI isn't part of the RFC
2616, and depends on the server side webserver stack. 4096 bytes appears
to be a good size based on apache, nginx, iis, tomcat defaults.

Fixes Unidata/netcdf-java#413


rschmunk:
FWIW, the CIS benchmarks for Apache security configuration recommend a `LimitRequestLine` value of 512. Or at least they did a few years ago. I haven't looked at the document recently. For a POST request, the recommended `LimitRequestBody` was 102400.

I've no idea how many server admins adhere to these recommendations.

PR #459, opened 2020-08-28
A few minor tweaks to our gradle pubs

lesserwhirls:
I ran into a few issues with our maven publications when migrating
threddsIso to use the new netcdf-java-bom artifact. Two things:

1. Make sure httpcore is included in the bom (needs to be matched to
   the version used by httpclient)
2. When using gretty, the runtime classpath of the project is modified,
   and this messes with what ends up in the pom (e.g. httpservices does
   not need a runtime dependency on jetty...). We were already modifying
   a pom to remove the inserted dependencies. However, now we modify the
   pom of projects using gretty as opposed to publishing a new artifact
   with the modified pom. Oops.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/459)

### DIFF
--- a/gradle/any/gretty.gradle
+++ b/gradle/any/gretty.gradle
@@ -76,7 +76,7 @@ if (project in publishedProjects) {
   // ain't nobody got time for that
   publishing {
     publications {
-      mavenJava(MavenPublication) {
+      mavenSources(MavenPublication) {
         pom.withXml {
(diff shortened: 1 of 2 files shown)


haileyajohnson (approved):
Well this one was easy.

lesserwhirls:
Thanks @haileyajohnson!

PR #464, opened 2020-09-02
Add cdmr client/server/iosp for CdmRemote over gRpc.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/464)

### DIFF
--- a/bufr/src/main/java/ucar/nc2/iosp/bufr/MessageUncompressedDataReader.java
+++ b/bufr/src/main/java/ucar/nc2/iosp/bufr/MessageUncompressedDataReader.java
@@ -321,9 +321,6 @@ public class MessageUncompressedDataReader {
 
     if (req.wantRow()) {
       Sequence seq = seqdd.refersTo;
-      if (seq == null) {
-        System.out.printf("HEY%n");
-      }
       assert seq != null;
 
       // for the obs structure
--- a/cdm/core/src/main/java/thredds/client/catalog/ServiceType.java
+++ b/cdm/core/src/main/java/thredds/client/catalog/ServiceType.java
@@ -4,15 +4,11 @@
  */
 package thredds.client.catalog;
 
-/**
- * Service Type enums
- *
- * @author caron
- * @since 1/7/2015
- */
+/** Service Type enums */
 public enum ServiceType {
   ADDE, // not used
   Catalog("Provide subsetting and HTML conversion services for THREDDS catalogs.", AccessType.Catalog, null), //
+  Cdmr("Provides index subsetting on remote CDM datasets, using cdmremote protocol.", AccessType.DataAccess, "cdmr"), //
   CdmRemote("Provides index subsetting on remote CDM datasets, using ncstream.", AccessType.DataAccess, "cdmremote"), //
   CdmrFeature("Provides coordinate subsetting on remote CDM Feature Datasets, using ncstream.", AccessType.DataAccess,
       "cdmrFeature"), //
--- a/cdm/core/src/main/java/ucar/nc2/ParsedSectionSpec.java
+++ b/cdm/core/src/main/java/ucar/nc2/ParsedSectionSpec.java
@@ -14,35 +14,30 @@ import java.util.List;
 
 /**
  * A String expression for denoting a section of a Variable to be read.
- * Allows nesting for subsetting of Structure member variables.
+ * Parse a section specification String. These have the form:
  *
- * @author caron
- * @since May 8, 2008
- * @see <a href=
- *      "https://www.unidata.ucar.edu/software/netcdf-java/reference/SectionSpecification.html">SectionSpecification</a>
+ * <pre>
+ *  section specification := selector | selector '.' selector
+ *  selector := varName ['(' dims ')']
+ *  varName := ESCAPED_STRING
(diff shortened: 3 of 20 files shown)


JohnLCaron:
ready

lesserwhirls:
I snuck in a gradle publication fix as well as merged in the dap4 PR - would you mind rebasing this one?

JohnLCaron:
ready

lesserwhirls (approved):
> Reviewed 19 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/464)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnLCaron)

___
*[cdmr/build.gradle, line 17 at r1](https://reviewable.io/reviews/unidata/netcdf-java/464#-MGJEoyMDvVHcDrHmJ4l:-MGJEoyN9D0i3iSoj3kD:byb29uj) ([raw file](https://github.com/unidata/netcdf-java/blob/f9f018a2558efdf3ec11f493b7936e7c3eb044f8/cdmr/build.gradle#L17)):*
> ```Gradle
> // Feel free to delete the comment at the next line. It is just for safely
> // updating the version in our release process.
> def grpcVersion = '1.31.1' // CURRENT_GRPC_VERSION
> ```

You could add this to  `gradle/any/shared-mvn-coords.gradle`

https://github.com/Unidata/netcdf-java/blob/6e8bb8eff22e19cb14b74f6b9f6ffae2b73b70e9/gradle/any/shared-mvn-coords.gradle#L21-L23

and then use something like `io.grpc:grpc-netty-shaded:${depVersion.grpc}` (to be defined) and `${depVersion.protobuf}` (already defined) so that the grpc and protobuf versions are defined in one place (unless protobuf 3.12.0 is required, as we are currently on 3.12.4).

I can do this, but I want to make sure that there isn't a strong dependency on protobuf 3.12.0.
___
*[cdmr/src/main/java/ucar/cdmr/CdmToProto.java, line 33 at r1](https://reviewable.io/reviews/unidata/netcdf-java/464#-MGJIKDg2kYcHNd8qEQ2:-MGJIKDg2kYcHNd8qEQ3:b3axe75) ([raw file](https://github.com/unidata/netcdf-java/blob/f9f018a2558efdf3ec11f493b7936e7c3eb044f8/cdmr/src/main/java/ucar/cdmr/CdmToProto.java#L33)):*
> ```Java
> 
> /** Convert between CdmRemote Protos and Netcdf objects. */
> public class CdmToProto {
> ```

When we use `apply from: "$rootDir/gradle/any/protobuf.gradle"` in `cdmr/build.gradle`, we modify the jacoco coverage reports to exclude `'**/*Proto.java'` and associated classes, to keep generated code out of our coverage reports. If we rename this class to something like `CdmToProtobuf`, then we'll get the coverage metrics.  
___
*[cdmr/src/main/proto/cdmRemote.proto, line 1 at r1](https://reviewable.io/reviews/unidata/netcdf-java/464#-MGJOWF0-xBy9p-8jQvW:-MGJOWF1ErgqK3gllewr:b-y075s1) ([raw file](https://github.com/unidata/netcdf-java/blob/f9f018a2558efdf3ec11f493b7936e7c3eb044f8/cdmr/src/main/proto/cdmRemote.proto#L1)):*
> ```protobuf
> syntax = "proto3";
> ```

For some reason I thought the proto directory structure had to match the package path (even though it can be defined in the proto itself), but I see that's not the case. (I think I was just following the example in the protobuf-gradle-plugin repo: https://github.com/google/protobuf-gradle-plugin/tree/master/examples/exampleProject/src/main/proto). What do you think? Should we try to standardize one way or the other?

JohnLCaron:
1. No dependence on protobuf 3.12.0, I was just cuta nd pasting from gRPC docs. Go ahead and clean it up when you get a chance.
2. Changed CdmToProto  -> CdmToProtobuf
3. I dont have much of an opinion; in a way the place where the code gets generated is arbitrary, and its nice to not have to negotiate the class heirarchy. Ill leave it to you.

Speaking of which, where is the task where the proto compiler runs?

lesserwhirls:
> 1. No dependence on protobuf 3.12.0, I was just cuta nd pasting from gRPC docs. Go ahead and clean it up when you get a chance.

On it.

> 2. Changed CdmToProto  -> CdmToProtobuf

👍 

> 3. I dont have much of an opinion; in a way the place where the code gets generated is arbitrary, and its nice to not have to negotiate the class hierarchy. Ill leave it to you.

It is nice to not have to negotiate the class hierarchy - let's go with a flat layout.

> Speaking of which, where is the task where the proto compiler runs?

When we apply the protobuf plugin, we get a few new proto tasks that run prior to the `compileJava` task:

~~~bash
:cdm:cdm-core:extractIncludeProto
:cdm:cdm-core:extractProto
:cdm:cdm-core:generateProto
:cdm:cdm-core:compileJava
~~~

The first two deal with extracting proto files from dependencies (either attached to the `compile` (first) or `protobuf` (second) configurations), and the third is where the actual proto compiler runs against the extracted proto files from the dependencies, as well as the proto files in `src/main/proto`. If you just wanted to compile the proto files, you could run the `generateProto` task.

PR #475, opened 2020-09-11
cdm-lite infrastructure work

lesserwhirls:
This PR does some work to `cdm-lite` to enable running a server, both through IntelliJ and using gradle + gretty. Details on how to do that below. A few things were needed to do this:

* Disable `@Autowire` of beans
 * `tds.properties` not included in the `cdm-lite` project, so could not do the `@Autowire` magic. However, we should think about the role of a `tds.properties` in a `cdm-lite`, and maybe even get rid of it all together?
* Fixed a typo in the name of the application context (was `app-context.xml`, but is really called `app-Context.xml`).
* Not critical for getting a server up-and-running, but replaced `@PostConstruct` with Spring's `InitializingBean` (if needed down the road).

Finally, this adds the machinery to have gradle spin up a local `cdm-lite` server prior to running the `cdm-lite` tests. What we should do is add a new source set, say `cdm-lite/fullServerTests`, in which all tests that need a local server spun up will live (and things like unit tests or mock tests can live in `cdm-lite/test`). That way, if we want to run a mock test, we don't have to wait for gretty to spin up a server that ultimately goes unused. Once we have a few tests going, I can come back and do the gradle magic to make that happen. Side note: this is something we should think about for other subprojects as well, like the `opendap` client and `httpservices`.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/475)

### DIFF
--- a/cdm-lite/build.gradle
+++ b/cdm-lite/build.gradle
@@ -15,8 +15,6 @@ dependencies {
   api project(':httpservices')
   api project(':opendap')
 
-  implementation 'com.google.guava:guava'
-
   compile 'com.google.guava:guava'
   compile 'org.jdom:jdom2'
   compile 'com.google.code.findbugs:jsr305'
@@ -67,4 +65,24 @@ war {
     into 'docs'
     expand properties
   }
+}
+
+// make the cdm-lite server available for tests
+import org.akhikhl.gretty.AppBeforeIntegrationTestTask
+import org.akhikhl.gretty.AppAfterIntegrationTestTask
+
+task('AppWarBeforeIntegrationTest', type: AppBeforeIntegrationTestTask) {
+  inplace = false
+  integrationTestTask 'test'
+}
+task('AppWarAfterIntegrationTest', type: AppAfterIntegrationTestTask) {
+  integrationTestTask 'test'
+}
+
+AppWarBeforeIntegrationTest.dependsOn assemble
+AppWarAfterIntegrationTest.mustRunAfter AppWarBeforeIntegrationTest
+
+gretty {
+  httpPort = 8080
+  contextPath = '/'
 }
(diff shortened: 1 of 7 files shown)


lesserwhirls:
# Using the local server from gradle

From the command line, run

`./gradlew :cdm-lite:appRunWar`

A local `cdm-lite` server will be accessible from `http://localhost:8080/`. Note that we don't have any exciting end points setup yet, but you can test that things are running by hitting `http://localhost:8080/tds/dods/hi.txt` (should get a 400 response, which is coming from the servlet and not the servlet container).

When you are finished, make sure to shut down the gretty controlled server properly. To do so, follow the instructions on the command line (i.e. hit any key and `return` to trigger the shutdown). If you forget and accidentally use `Ctrl+c` or something like that, the process won't release the port. The next time you try to startup the server, you will get a "port in use" error. Boo 👎  To fix this, run `./gradle --stop`, which doesn't actually stop the server directly, but stops the underlying gradle daemon, which will result in what we want.

# Using the local server from IntelliJ

This one is a little tricky, but not terrible. IntelliJ has issues with the generated proto code, and gets confused, so we have to do a little work to bypass IntelliJ's internal build process. Here are the steps to add a Tomcat run/debug configuration to IntelliJ:

1. Run the assemble gradle task (from the command line, simply `./gradlew assemble`) - this will make sure the cdm-lite war file exists, which will make things easier in a few steps.
1. Through the IntelliJ interface, add a new Run/Debug configuration
new-config
1. ...and add a new Local Tomcat configuration
new-tomcat-config
1. Give the configuration a name, switch to the `deployment` tab, and add a new External Source for deployment at server startup
config-deploy-0
1. Choose the `cdm-lite` war artifact that was assembled in step 1.
config-deploy-1
1. Add a new "Run Gradle Task" to the "Before launch" configuration
config-deploy-2
1. Choose the `:cdm-lite` project and use `assemble` as the task
config-deploy-3
1. Finally, remove the `Build` task from the list of tasks from the "Before launch" options. The end result should look like this:
config-deploy-4

Save the configuration. w00t. Now you should be able to run the `cdm-lite` server in normal or debug mode through IntelliJ.

final

JohnLCaron:
This looks good. I havent yet actually tested it, but i will asap. You can check it in if you want.

The docs a re awesome, thanks so much!

PR #451, opened 2020-08-19
Cleanup nc2.dataset, part 1

JohnLCaron:
Move CoordTransformFactory, StructurePseudoDS and StructurePseudoDS2D to internal.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/451)

PR #478, opened 2020-09-12
Netcdf4-jna port to builder framework (part 1).

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/478)


JohnLCaron:
ready

lesserwhirls (approved):
> Reviewed 60 of 60 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/478)* status: :shipit: complete! all files reviewed, all discussions resolved

lesserwhirls:
Thanks @JohnLCaron! There are a few places where you explicitly call out the use of C++ (netCDF or HDF), but we're actually wrapping the C library calls. Minor though. You could snag those before a rebase, or just go for the rebase at this point - either way is good with me.

There are still a few things under the `ucar.nc2.jni.netcdf` that could probably be relocated to other packages, or to an internal package, but that's something to think about later I think. I think netCDF-Java v6 is where I'd like to start writing ` module-info.java` files so that those using >Java 9 will be able to cleanly use the Java Platform Module System (Java 8 will just ignore them). If most non-public packages live under an external package name, it could assist those moving to use JPSM when netCDF-Java starts using Java 11 features.

JohnLCaron:
Not sure why we get this failure, it succeeds locally:

ucar.nc2.TestSequence > readStructureWithinSequence FAILED
2020-09-14T22:17:34.8923205Z     java.io.IOException: java.lang.RuntimeException: resource not found=</resources/bufrTables/local/tableLookup.csv>

C++ -> C in next PR

lesserwhirls:
The test passes for me locally as well (running it as an individual test), but I am able to get it to fail if I run `./gradlew testAll`, so that's a clue. I have a few ideas and am looking into it.

lesserwhirls:
Ok...damn. That was a tough one to find. I'm not 100% sure what is up, but I know that the test will pass if I make a change to the `cdm-test` subproject's [`TestRuntimeConfigParser`](https://github.com/Unidata/netcdf-java/pull/481/commits/17a4faf5f6d83113830a64aa561d06753533eb93) test such that the test config points to a valid bufr lookup table.

One of two possibilities.

1. The changes made in Unidata/netcdf-java#451, specifically to the [RuntimeConfigParser](https://github.com/Unidata/netcdf-java/pull/451/commits/c441cc5a8aeb51c800cbe8dda9041069b89754eb#diff-6614671b44c4d787a8f5248ffa8edcf6L230-L262), made missing lookup tables less tolerable. If that's the case, I would have expected things to fail before this PR.

2. The order of the tests changed with all of the gradle subproject shuffling (like the addition of the `netcdf4` subproject) (apparently gradle [does not have a reproducible test order](https://github.com/gradle/gradle/issues/8520)), and I believe we've been hit with this before. If this is the case, then I think the underlying issue is that the test `cdm-test/src/test/java/ucar/nc2/internal/util/xml/TestRuntimeConfigParser.java` is now running *before* `ucar.nc2.TestSequence`, which adds a non-existent bufr lookup table, which then causes the sequence test to barf when the bufr IOSP tries to load it.

Maybe it's a combination of both. At any rate, if you change that one line in the `cdm-test` subproject's [`TestRuntimeConfigParser`](https://github.com/Unidata/netcdf-java/pull/481/commits/17a4faf5f6d83113830a64aa561d06753533eb93) test, then we should be good to go.

JohnLCaron:
Thanks for tracking the problem down. Im noticing how I give you the hard stuff. 

Im submitting a separate fix for it.

lesserwhirls:
With Unidata/netcdf-java#485 in, I think these will go quick with rebases. Shall we start with this one?

JohnLCaron:
ready

lesserwhirls (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/478)* status: :shipit: complete! all files reviewed, all discussions resolved

lesserwhirls:
Sorry that took so long to get in! Thank you @JohnLCaron!

PR #483, opened 2020-09-18
create ignore tag got include-code-block plugin; add tests for netcdf

haileyajohnson:
1) create ignore tag (/*DOCS-IGNORE*/) for include-code-block plugin
2) move CDM dataset tutorial code to java functions and update (including updating Tools UI screenshots)
3) add minimal tests for CDM dataset tutorial code
4) remove native HTML from docs

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/483)

PR #493, opened 2020-09-24
port #483

haileyajohnson:
tutorials; remove native HTML from docs

fix test failure from invalid range

rewrite write netcdf page; move list of modules to cdm file type page

incorporate PR suggestions

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/493)


lesserwhirls (approved):
> Reviewed 21 of 21 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/493)* status: :shipit: complete! all files reviewed, all discussions resolved

lesserwhirls:
Not as bad as I thought! The `DiskCache` import with `internal` in its package gives me pause, but v6 is very much a work in progress, so it remains to be seen how that will all shake out in the end. Once nice thing here is that as things move around in 6, we will get some real-time feedback based on the tutorial materials. Thank you @haileyajohnson!

PR #479, opened 2020-09-14
Refactor Netcdf writing (pass 1)

JohnLCaron:
NetcdfFormatWriter only creates, doesnt update.
NetcdfFormatUpdater does the updating (not finished).
Remove IOServiceProviderWriter.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/479)

Issue #352, opened 2020-07-07
netcdf4 user defined types not handled correctly


JohnLCaron:
See TestNc4EnumWriting as an example.

JohnLCaron:
Problem with enums:

cdl:

netcdf writeEnumFromCdl {
  types:
    short enum dessertType_t {dirt = 0, pie = 18, donut = 268, cake = 3284};
  dimensions:
    time = UNLIMITED;
  variables:
    dessertType_t dessert(time);
}

ncgen -o C:/temp/writeEnumFromCdl.nc4 -k 3 C:/dev/github/netcdf-java/netcdf4/src/test/resources/ucar/nc2/jni/netcdf/enum.cdl

$ ncdump C:/temp/writeEnumFromCdl.nc4
netcdf C\:/temp/writeEnumFromCdl {
types:
  short enum dessertType_t {dirt = 0, pie = 18, donut = 268, cake = 3284} ;
dimensions:
        time = UNLIMITED ; // (0 currently)
variables:
        dessertType_t dessert(time) ;
data:
}

but looking at the hdf messages, dessert has a self-contained DataType message:

message type = Datatype(3);  datatype= 8 byteSize= 2 NCtype= null  flags=  4 0 0 endian= LITTLE enumTypeName= dessert 
parent base= { datatype= 0 byteSize= 2 NCtype= short  flags=  8 0 0 endian= LITTLE}

Note that enumTypeName= dessert, not dessertType_t !!

There is a seperate datatype message for the "dessertType_t" "user type":

message type = Datatype(3);  datatype= 8 byteSize= 2 NCtype= null  flags=  4 0 0 endian= LITTLE enumTypeName= dessertType_t parent base= { datatype= 0 byteSize= 2 NCtype= short  flags=  8 0 0 endian= LITTLE}

checking with h5dump:

$ h5dump C:/temp/writeEnumFromCdl.nc4
HDF5 "C:/temp/writeEnumFromCdl.nc4" {
GROUP "/" {
   ATTRIBUTE "_NCProperties" {
      DATATYPE  H5T_STRING {
         STRSIZE 34;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "version=2,netcdf=4.7.4,hdf5=1.10.6"
      }
   }
   DATASET "dessert" {
[...]

PR #496, opened 2020-09-26
Fix reading enums from netCDF-4 files

lesserwhirls:
Partial backport of Unidata/netcdf-java#479 (changes related to reading enums from netCDF-4 files). See Unidata/netcdf-java#352.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/496)

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/EnumTypedef.java
+++ b/cdm/core/src/main/java/ucar/nc2/EnumTypedef.java
@@ -175,6 +175,18 @@ public class EnumTypedef extends CDMNode {
 
   }
 
+  // Needed for netCDF4 wierdness
+  public boolean equalsMapOnly(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    EnumTypedef that = (EnumTypedef) o;
+    return com.google.common.base.Objects.equal(map, that.map);
+  }
+
   @Override
   public int hashCode() {
     String name = getShortName();
--- a/cdm/core/src/main/java/ucar/nc2/Group.java
+++ b/cdm/core/src/main/java/ucar/nc2/Group.java
@@ -924,7 +924,7 @@ public class Group extends CDMNode implements AttributeContainer {
     private NetcdfFile ncfile; // set by NetcdfFile.build()
     private AttributeContainerMutable attributes = new AttributeContainerMutable("");
     private List<Dimension> dimensions = new ArrayList<>();
-    private List<EnumTypedef> enumTypedefs = new ArrayList<>();
+    public List<EnumTypedef> enumTypedefs = new ArrayList<>();
     private boolean built;
 
     public Builder setParentGroup(@Nullable Group.Builder parentGroup) {
--- a/cdm/core/src/main/java/ucar/nc2/NetcdfFile.java
+++ b/cdm/core/src/main/java/ucar/nc2/NetcdfFile.java
@@ -135,6 +135,7 @@ public class NetcdfFile implements FileCacheable, Closeable {
   @Deprecated
   public static final String IOSP_MESSAGE_REMOVE_RECORD_STRUCTURE = "RemoveRecordStructure";
   public static final String IOSP_MESSAGE_RANDOM_ACCESS_FILE = "RandomAccessFile";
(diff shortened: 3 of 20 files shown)


lesserwhirls:
Ok, we're passing here and have no new failures on Jenkins. Whew!

haileyajohnson (commented):
> Reviewed 20 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/496)* status: all files reviewed, 2 unresolved discussions (waiting on @lesserwhirls)

___
*[cdm/core/src/main/java/ucar/nc2/Group.java, line 927 at r1](https://reviewable.io/reviews/unidata/netcdf-java/496#-MIKd_zm45dLzxMfhG0B:-MIKd_zm45dLzxMfhG0C:b-t88v1v) ([raw file](https://github.com/unidata/netcdf-java/blob/34dfa90c9dba2882e95bc4f9069dea6849e622d2/cdm/core/src/main/java/ucar/nc2/Group.java#L927)):*
> ```Java
>     private AttributeContainerMutable attributes = new AttributeContainerMutable("");
>     private List<Dimension> dimensions = new ArrayList<>();
>     public List<EnumTypedef> enumTypedefs = new ArrayList<>();
> ```

I think I'm missing where this is used that it needs to be public...
___
*[cdm/core/src/main/java/ucar/nc2/iosp/netcdf3/N3iosp.java, line 1008 at r1](https://reviewable.io/reviews/unidata/netcdf-java/496#-MIKd7ke43IHXHAa4gNX:-MIKd7ke43IHXHAa4gNY:ba91hvp) ([raw file](https://github.com/unidata/netcdf-java/blob/34dfa90c9dba2882e95bc4f9069dea6849e622d2/cdm/core/src/main/java/ucar/nc2/iosp/netcdf3/N3iosp.java#L1008)):*
> ```Java
>   public String getFileTypeVersion() {
>     return "1";
>   }
> ```

can these strings be class constants?

lesserwhirls (commented):
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/496)* status: all files reviewed, 2 unresolved discussions (waiting on @haileyajohnson)

___
*[cdm/core/src/main/java/ucar/nc2/Group.java, line 927 at r1](https://reviewable.io/reviews/unidata/netcdf-java/496#-MIKd_zm45dLzxMfhG0B:-MILvCUE3V2b9_Ac3qLe:b-nfwwzq) ([raw file](https://github.com/unidata/netcdf-java/blob/34dfa90c9dba2882e95bc4f9069dea6849e622d2/cdm/core/src/main/java/ucar/nc2/Group.java#L927)):*

I think I'm missing where this is used that it needs to be public...
</blockquote></details>

The access is down in `cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java`, here:

https://github.com/Unidata/netcdf-java/pull/496/files#diff-552511b78a00abd0c3ead642c8491504R1573

There are cases where the enum name -> value map is attached to one object in the HDF5 file, but isn't named (even when there might be a named enum object with the same name -> value map). More info over at https://github.com/Unidata/netcdf-java/issues/352. In this case, we are streaming through the enums that have been added to the Group builder by name, and looking for any that have the same name -> value map. If we find one, we use it. rather than create a new one. If the group object was already built, we could use the `getEnumTypedefs()` method, but since we're still in the process of building the `NetcdfFile` object, we need to see what has been added to the builder up to this point.
___
*[cdm/core/src/main/java/ucar/nc2/iosp/netcdf3/N3iosp.java, line 1008 at r1](https://reviewable.io/reviews/unidata/netcdf-java/496#-MIKd7ke43IHXHAa4gNX:-MIM-1AAC9mGCt-01U3q:b1g8chq) ([raw file](https://github.com/unidata/netcdf-java/blob/34dfa90c9dba2882e95bc4f9069dea6849e622d2/cdm/core/src/main/java/ucar/nc2/iosp/netcdf3/N3iosp.java#L1008)):*

can these strings be class constants?
</blockquote></details>

Yep, and we could even use `String.valueOf(NetcdfFileFormat.NETCDF3.version)` since it's already defined there! Let me fix those up real quick.

haileyajohnson (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/496)* status: all files reviewed, 2 unresolved discussions (waiting on @haileyajohnson)

PR #497, opened 2020-09-26
Fix Vlens,  iterator returns Array<T>

JohnLCaron:
Fix Vlens. Currently only used by H5iosp. 
Change Vlen to use StorageMutable<Array<T>> not StorageMutable<T[]>, so the iterator returns Array<T>.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/497)


JohnLCaron:
ready

Issue #346, opened 2020-07-02
NetCdfDataset.openDataset doesn't work for local OpeNDAP/DODS Files


furlong-tw:
When working with a set of OpeNDAP/DODS files (.dods, .dds, .das) stored locally the general NetCdfDataset.openDataset method is not capable of of parsing the data.

### Steps to reproduce

- Using version 5.3.3
- Setup a data folder '/home/user/data' that contains wna20200629_00z.dods, wna20200629_00z.das, wna20200629_00z.dds files
- Call NetCdfDataset.openDataset with the path to the OpeNDAP/DODS files.

### Current behavior

Two possible options exist

1. Pass the base path to the files, example: "file:/home/user/data/wna20200629_00z". Results in "No such file or directory"
2. Pass the path to the primary data file, example:  "file:/home/user/data/wna20200629_00z.dods". Results in "File not readable wna20200629_00z.dods.dds".

### Expected Behavior

Calling:
 `NetCdfDataset.openDataset("file:/home/user/data/wna20200629_00z")` 
should produce the same result as:
` new DODSNetcdfFile( "file:/home/user/data/wna20200629_00z" )`

Directly instantiating a DODSNetCdfFile behaves as anticipated and ingests the local data file.

### Potential line of inquiry
The issue may be related to how arbitrary strings are correlated to DatasetUrl. If "file" protocol is passed then a file ending of ".dods", ".dds", or "das" is required to result in ServiceType.OPENDAP. If "dods" protocol is used then it is automatically modified to "https avoiding local file usage.

See [DatasetUrl](https://github.com/Unidata/netcdf-java/blob/7c2e46606fb2d2032c2612a900e7f6376c40bf42/cdm/core/src/main/java/ucar/nc2/dataset/DatasetUrl.java#L285):

```
  private static ServiceType decodePathExtension(String path) {
    // Look at the path extensions
    if (path.endsWith(".dds") || path.endsWith(".das") || path.endsWith(".dods"))
      return ServiceType.OPENDAP;

    if (path.endsWith(".dmr") || path.endsWith(".dap") || path.endsWith(".dsr"))
      return ServiceType.DAP4;

    if (path.endsWith(".xml") || path.endsWith(".ncml"))
      return ServiceType.NCML;
    return null;
  }
```

[...]

PR #499, opened 2020-09-28
Support datasets with multiple variables

scopeForImagination:
Specifically, correctly search for the requested variable of interest.

For example, if you download multiple variables using the * query
e.g. https://nomads.ncep.noaa.gov/dods/wave/wna/wna20200915/wna20200915_00z.dods?*

The parser currently can only return the first variable in the file.  With a minor code change, the code can correctly search for the variable requested by the caller.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/499)


lesserwhirls:
Greetings @scopeForImagination! As far as I understand, the wildcard expression isn't a valid  DAP2 constraint expression. I'm trying to figure out your specific use case to see if there is a better way of going about achieving your end goal. Are you using `DodsNetcdfFile` directly to access data from the dap endpoint? Ultimately, are you trying to read data into a netCDF-Java `Array` object for further processing, or are you trying to download the data and save it into a file on disk?

scopeForImagination:
Hi @lesserwhirls .  I can't seem to find documentation for constraint expressions at the moment, but if the concern was that the url example I provided doesn't work anymore, it's because the data has expired.  Try a more recent date, such as yesterday: https://nomads.ncep.noaa.gov/dods/wave/wna/wna20200929/wna20200929_00z.dods?*  and you'll see that it downloads all of the variables, versus a more specific query such as https://nomads.ncep.noaa.gov/dods/wave/wna/wna20200929/wna20200929_00z.dods?windsfc only downloads one variable.  Alternatively, you can list multiple variables: https://nomads.ncep.noaa.gov/dods/wave/wna/wna20200929/wna20200929_00z.dods?lat,lon,time,htsgwsfc,dirswsfc - doesn't use the wildcard, but has more than one variable. Or consider this query: http://test.opendap.org/dap/data/ff/gsodock.dat.html which doesn't include a wildcard, but is capable of requesting multiple variables.  Regarding the use case, we are using DodsNetcdfFile (specifically, DodsNetcdfFile netcdfFile = DodsNetcdfFile.builder().build(basePath, null);) to load it into a NetcdfDataset.  We'd like eventually to do both end goals of (a) reading the data into an array for further processing and (b) downloading the data and saving to a file, but our current use case is (a).

lesserwhirls:
Ok, I think I see better now what you are trying to do. But to be sure, are you 1) obtaining the binary response from the dap2 server outside of netCDF-Java and trying to read that in through the library, or 2) are you passing in the url into the library directly using something like `DODSNetcdfFile("https://nomads.ncep.noaa.gov/dods/wave/wna/wna20200915/wna20200915_00z.dods?*")`?

As a side note, some dap2 servers support wildcards, but others do not. Saying it's not valid was not correct on my part - servers can implement extensions to the constraint expressions, but I don't think the wildcard expression is one that is mandated by the dap2 spec.

scopeForImagination:
Option 1. We're using the url to manually download the binary to a file on disk and then loading the local file using DodsNetcdfFile (i.e. in my example code in my last comment, basePath is a path to a local file).

lesserwhirls:
Ah, ok! So going through `NetcdfDataset.open(...)` likely put you in the same situation as [this issue](https://github.com/Unidata/netcdf-java/issues/346) describes? Ideally that would work, and you would not need to go through `DODSNetcdfFile` directly...but even if `NetcdfDataset.open(...)` did the right thing, it looks like read would fail in this case since it does not search for the correct variable in the file. @JohnLCaron - did the opendap code previously support opening a file containing the binary response from a dap2 server? It sounds like we're really close to being able to do that (like maybe `DODSNetcdfFile` used to be able to work as a .dods file IOSP)?

scopeForImagination:
Correct - I get the same error as described in the linked issue when using NetcdfDataset.open(...).
And correct - the read would still fail because it would still always return the first variable in the file, regardless of which variable was requested by the caller.

scopeForImagination:
@lesserwhirls any updates?

lesserwhirls:
No updates. Let me run our full test suites on your PR just to be sure, and if everything looks ok, then we'll go head and get this merged in. It will take a little more work to get it working through the public API, but for now, at least you will have something. Once reading via `NetcdfFiles` and `NetcdfDataset` works, then you'll want to use those instead of `DODSNetcdfFile` directly.

scopeForImagination:
That sounds great - thanks, @lesserwhirls.

lesserwhirls:
The full tests pass and everything looked good. I took a deeper look at the code, and went ahead and fixed the more general problem such that you will be able read `.dods` files through the public API (e.g. `NetcdfFiles.open*`, `NetcdfDatasets.open*`), which also fixes Unidata/netcdf-java#346. I did that work on the `maint-5.x` branch, and will port it to the `develop` branch. If it is ok with you, when I port to the `develop` branch, I'll checkout your branch first and make my additions starting after your commit. Github will allow me to push those changes to your branch, or I can push them to a new branch on my fork to make a PR. Either way works for me - I just don't want to stomp on any further work you might have done or will be doing on your branch. Thank you @scopeForImagination!

scopeForImagination:
Fantastic news!  Thanks @lesserwhirls  - I am finished with that branch, so feel free to do whatever is easiest.

lesserwhirls:
Ok, we have this in place on both branches (should be available in the 5.4.0-SNAPSHOT artifacts). Details on how it works can be found at https://docs.unidata.ucar.edu/netcdf-java/5.4/userguide/dataset_urls.html#opendap-datasets. Note that if you want the metadata associated with the dataset (which would show up as attributes), you will need to also have the `.das` response saved to a file (using your example from above, you would save  https://nomads.ncep.noaa.gov/dods/wave/wna/wna20200915/wna20200915_00z.das in addition to the .dods response). Thank you once again!

PR #501, opened 2020-09-29
Add gradle dependency-check plugin

lesserwhirls:
Port of Unidata/netcdf-java#500.

The OWAPS dependency-check plugin for Jenkins is up for adoption, and
given that it has been failing to run for a week or so, does not seem to
be the way to go forward. This PR adds a gradle plugin as a replacement.
We will continue to use Jenkins to run the check on a regular schedule,
but now using the gradle machinery in place of the jenkins plugin.

To run the check locally, use:

./gradlew dependencyCheckAggregate

The plugin will generate an html report which can be found at

build/reports/dependency-check-report.html

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/501)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ Thumbs.db
 # Coverity files 
 cov-int
 thredds.tgz
+
+# dependency check NVD files
+project-files/owasp-dependency-check/nvd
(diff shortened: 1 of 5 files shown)


haileyajohnson (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/501)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #502, opened 2020-09-30
Make sure projection setters also set base value

lesserwhirls:
When backporting Unidata/netcdf-java#367 in Unidata/netcdf-java#370, I
missed updating the setter methods (Projections are immutable in version
6). Also, found a case where we were still relying on a calculated value
for the scale parameter in the constructCopy of Sterographic, so fixed that
as well (will need to port that to the develop branch).

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/502)


haileyajohnson (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/502#-:-MIZjMmj1Nt1DmPa96Kx:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #507, opened 2020-10-01
add new DataViewer service type

haileyajohnson:
Add new service type for services with are data viewers (e.g. the Jupyter Notebook service)
[tds#116](https://github.com/Unidata/tds/pull/116) depends on this change

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/507)


dopplershift:
@lesserwhirls is this another of those things that doesn't make great sense to have in netcdf-java vs. tds?

PR #508, opened 2020-10-02
Cmdr fixes

JohnLCaron:
Make cmdr work with nested Structures, Sequences.
Chars always use one byte in cmdr encoding.
Rework StructueMembers.setStandardOffsets().
StructureDataStorageBB nested StructureArray was wrong, need to track offset ourself.
Extensively rework CmdrConverter, esp for nested StructureData.
Add StructureMembersProto to cmdr protos to make recursion easier.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/508)


JohnLCaron:
ready

PR #509, opened 2020-10-03
Fix Bufr missing value overflow, compressed reader.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/509)


JohnLCaron:
ready

JohnLCaron:
See PR#210. Not sure what the original problem was; the original file seems ok:
TestDir.cdmUnitTestDir + "formats/bufr/userExamples/problems/0-01-030_bitWidth-128.bufr

lesserwhirls (approved):
> Reviewed 41 of 41 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/509)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnLCaron)

___
*[cdm/core/src/main/java/ucar/nc2/internal/util/CompareArrayToMa2.java, line 6 at r1](https://reviewable.io/reviews/unidata/netcdf-java/509#-MIs_xts0IEGoedZA5tn:-MIs_xtt7oFF7RFo9IEI:bvn8d76) ([raw file](https://github.com/unidata/netcdf-java/blob/9c455925194158e4ce4433a84938b4cf40fdf94b/cdm/core/src/main/java/ucar/nc2/internal/util/CompareArrayToMa2.java#L6)):*
> ```Java
>  */
> 
> package ucar.nc2.internal.util;
> ```

Could be a good candidate to go into the `cdm-test-utils` subproject.
___
*[cdmr/build.gradle, line 13 at r1](https://reviewable.io/reviews/unidata/netcdf-java/509#-MIsak2dCutId5VbtY7k:-MIsak2dCutId5VbtY7l:b-k69xry) ([raw file](https://github.com/unidata/netcdf-java/blob/9c455925194158e4ce4433a84938b4cf40fdf94b/cdmr/build.gradle#L13)):*
> ```Gradle
>   api project(':cdm:cdm-core')
> 
>   // for CdmrServer: we should just add on the classpath?
> ```

Since we're using the `java-library` gradle plugin, we have a few options:

`api` is for dependencies that are exposed via the public api.
`implementation` is for dependencies that are only used internally.

Gradle see both of those as compile level dependencies when we are building he project, but users of the project would see `api` as something they need to have at compile time, and `implementation` as something needed at runtime.

If we don't actually call anything from `bufr` or `grib` directly anywhere `cdmr`, then we can use `runtimeOnly`, which makes it a runtime dependency for both us and users of the `cdmr` project.

lesserwhirls:
Two minor comments, but nothing big. I'll get this one in, which will also take care of Unidata/netcdf-java#508

dopplershift:
I *think* if you've added "hacktoberfest" to the repository topics it should be enough just to merge the PR to have it count.

PR #512, opened 2020-10-07
Fix reading of local .dods files

lesserwhirls:
Add support for reading `.dods` files (binary response from a dap2 server saved in a file) through the public API. ~The Dataset Descriptor Structure (`.dds`) response **must** be present in a separate file (in the same location with the same base file name, but with a `.dds` extension)~. A ~third~ second file containing the Data Attribute Structure (`.das`) response ((in the same location with the same base file name, but with a `.das` extension) is optional, but _highly encouraged_ as it contains dataset metadata.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/512)


ethanrd:
I think the .dods response contains the .dds response (of what is in the .dods, not the full dataset). Does DAP2 support a subset request for .das? Or is the .das for the full dataset required?

lesserwhirls:
Thanks for the feedback and questions Ethan! So yeah, the DDS is contained within the `.dods` response, per the dap2 spec, so I've removed the need to have a separate `.dds` file.

According to the [DAP2.0 spec](https://www.opendap.org/pdf/dap_2_data_model.pdf) section on _Constraint Expressions_:

> The constraint expression MUST be encoded using US-ASCII characters. It MAY be used when requesting the
DDS or DataDDS (i.e. when using the dds or dods extensions, see Section 7.2.3 on page 30). It MAY NOT be
used with the DAS, Version or Help Requests.

so it does not seem like one should expect to be able to subset the `.das` (our server ignores any CE on the `.das` request).

ethanrd:
Oops! Sorry about the change to "ready for review".

ethanrd:
@lesserwhirls - That all sounds good.

lesserwhirls:
No new failures on Jenkins - I think we're ready for review now. @ethanrd @haileyajohnson - would one of you mind giving it one more look over?

haileyajohnson (approved):
> Reviewed 7 of 13 files at r1, 6 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/512)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ethanrd)

ethanrd (approved):
Looks good.

Issue #532, opened 2020-10-22
EOFException on NCDump data


Mikhail-Aoun:
Hi,
As explained in [https://github.com/Unidata/thredds/issues/1342](https://github.com/Unidata/thredds/issues/1342), we've found an issue when trying to read data from some particular netcdf files. 
The same bug happens with Tools UI latest stable release (Version 5.3.3) and the same netcdf files as the one attached in the other issue. This time, we met an EOFException in the ucar.nc2.internal.iosp.hdf5.H5objects$GlobalHeap instantiation algorithm.

The Tools UI Exception when you open this file and try to NCDump tm_packets data is : 

> ubyte tm_packets(number_tm_packets=10256, *);
>   :long_name = "Array of TM packets";
> 
>  data:
> java.io.EOFException
> 	at ucar.unidata.io.RandomAccessFile.readInt(RandomAccessFile.java:1233)
> 	at ucar.unidata.io.RandomAccessFile.readLong(RandomAccessFile.java:1319)
> 	at ucar.nc2.internal.iosp.hdf5.H5headerNew.readLength(H5headerNew.java:2215)
> 	at ucar.nc2.internal.iosp.hdf5.H5objects$GlobalHeap.<init>(H5objects.java:2631)
> 	at ucar.nc2.internal.iosp.hdf5.H5objects$HeapIdentifier.getHeapObject(H5objects.java:2544)
> 	at ucar.nc2.internal.iosp.hdf5.H5headerNew.getHeapDataArray(H5headerNew.java:2055)
> 	at ucar.nc2.internal.iosp.hdf5.H5headerNew.getHeapDataArray(H5headerNew.java:2048)
> 	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:338)
> 	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:255)
> 	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:191)
> 	at ucar.nc2.NetcdfFile.readData(NetcdfFile.java:2110)
> 	at ucar.nc2.Variable.reallyRead(Variable.java:761)
> 	at ucar.nc2.Variable._read(Variable.java:731)
> 	at ucar.nc2.Variable._read(Variable.java:773)
> 	at ucar.nc2.Variable.read(Variable.java:597)
> 	at ucar.nc2.NetcdfFile.readSection(NetcdfFile.java:2139)
[...]

PR #533, opened 2020-10-22
Fix for EOFException on NCDump data Unidata/netcdf-java#532

Mikhail-Aoun:
On GlobalHeap instantiation, we should stop reading file data just after we met a HeapObject.id equals to 0. Otherwise, we could be at the end of the file and a EOFException will be raised. Addresses the github issue Unidata/netcdf-java#532.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/533)


lesserwhirls:
Thank you for opening an issue, supplying a test file, and providing a fix! These kinds of bugs can be tricky to track down, especially without a sample file to trigger it (which we didn't have before, but now that we do it will get tested!). Very much appreciated!

Mikhail-Aoun:
You're welcome.
Thanks for the tests and code review !

PR #549, opened 2020-11-06
Cleanup CoordinateSystems, Part 2.

JohnLCaron:
Make classes immutable where possible.
isComplete() excludes dimensions of length < 2.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/549)


JohnLCaron:
ready

lesserwhirls (approved):
I noticed a few things noted as "technically not immutable" - are those in place until 7, or just a necessary evil internally at the moment?

> Reviewed 32 of 32 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/549#-:-MLV1YOB3T2NpH5ZQPlp:b2r8g42)* status: :shipit: complete! all files reviewed, all discussions resolved

JohnLCaron:
Its possible that they can be made immutable, and ill have another pass when the old dt and ft classes are removed. I had immutable solutions, but it was interfering with easy unit testing. So I realized that strict immutablity is only one goal that has to be balanced against others.

The solution I have now still keeps most of those classes practically immutable. The really nice thing about strict immutability is that the compiler can give warnings, and in the case of final fields, enforce immutability.

PR #564, opened 2020-11-17
Deal with Grib2 percentile variables.

JohnLCaron:
You probably want to backport this to 5 and maybe 4.

It affects tds/NCEP/NBM/Ocean/*, which is currently failing when trying to open, because there are multiple variables with the same name, because the name does not include the percentile. You want to delete all ncx4 files and allow to be recreated with new code.

Problem also shows up as having duplicate records. The new Iosp > GRIB > CdmIndexScan widget in UI makes it easier to see those problems.

Im not sure if other TDS datasets are affected, at the moment I dont think so.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/564)


JohnLCaron:
ready

Issue #433, opened 2020-08-14
Handling large dataset from S3 fails.


hyoklee:
* cdm-core-5.4.0-SNAPSHOT.jar:5.4.0-SNAPSHOT / TDS 5.0.0 beta8
* Linux ip-172-31-25-188 5.3.0-1032-aws #34~18.04.2-Ubuntu SMP Fri Jul 24 10:06:28 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
* openjdk version "11.0.8" 2020-07-14 OpenJDK Runtime Environment (build 11.0.8+10-post-Ubuntu-0ubuntu118.04.1) OpenJDK 64-Bit Server VM (build 11.0.8+10-post-Ubuntu-0ubuntu118.04.1, mixed mode, sharing)
*  Issue: 
```
java.lang.StackOverflowError: null
ucar.unidata.io.RemoteRandomAccessFile.readFromCache(Remote RandomAccessFile.java:107
```
* Steps to reproduce:
  1. Run Ubuntu EC2 on AWS us-west-2.
  2. Run TDS and create a catalog for [Terra Fusion Data Sampler](https://registry.opendata.aws/terrafusion/) s3://terrafusiondatasampler/P108/TERRA_BF_L1B_O10204_20011118010522_F000_V001.h5
  3. Try to read a large dataset (e.g., /MODIS/granule_2001322_0110/_1KM/Geolocation/Longitude via OPeNDAP web interface or pydap/netCDF4-python.
  4. Try to export or dump the above dataset using toolsUI.

pnunu:
- cdm-core 5.3.3
- windows 10  jdk1.8

- issue
Exception in thread "main" java.lang.StackOverflowError
	at ucar.unidata.io.RemoteRandomAccessFile.readFromCache(RemoteRandomAccessFile.java:99)
	at ucar.unidata.io.RemoteRandomAccessFile.readFromCache(RemoteRandomAccessFile.java:106)

lesserwhirls:
I've been trying to reproduce this using the latest code, but so far I have not had any luck. The location where the overflow is happening indicates that we are likely in an infinite regression situation, which is not good (I'm going to make a change to better handle that case).

[...]

PR #568, opened 2020-11-19
Improve S3 access and NcML Aggregation support by Coverage API

lesserwhirls:
This PR does a few things:
 * Improve Coverage support for NcML aggregations (which produce Time coordinate variables of type String)
 * Remove regression from S3 read logic (fixes Unidata/netcdf-java#433)
 * Improve S3 read testing

In addition, this PR also knocks out a few test errors encountered on Jenkins. These errors are not code related (post-test file cleanup too broad, files in test datasets moved).

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/568)

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/grib/TestGribIndexCreation.java
+++ b/cdm-test/src/test/java/ucar/nc2/grib/TestGribIndexCreation.java
@@ -3,6 +3,7 @@ package ucar.nc2.grib;
 import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.slf4j.Logger;
@@ -350,6 +351,8 @@ public class TestGribIndexCreation {
 
   @Test
   @Category(NeedsCdmUnitTest.class)
+  @Ignore("There appears to be an issue with the encoding of these messages, so they have been moved to "
+      + "cdmUnitTest/gribCollections/ecmwf/exclude/bcs/ until we can figure out what ECMWF has done to make these.")
   public void createECMWFbcs() throws IOException { // SRC
     FeatureCollectionConfig config =
         new FeatureCollectionConfig("ECMWFbcs", "test/ECMWFbcs", FeatureCollectionType.GRIB1,
--- a/cdm/core/src/main/java/ucar/nc2/dataset/CoordinateAxis1DTime.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/CoordinateAxis1DTime.java
@@ -201,22 +201,15 @@ public class CoordinateAxis1DTime extends CoordinateAxis1D {
 
   @Override
   protected void readValues() {
-    // if DataType is not numeric, handle special
-    if (!this.dataType.isNumeric()) {
+    // if orgVar DataType is not numeric (e.g. Char or String), read from the cdates array that was created when
+    // the axis was created by this classes factory.
+    if (this.orgDataType != null && !this.orgDataType.isNumeric()) {
       this.coords = cdates.stream().mapToDouble(cdate -> (double) cdate.getDifferenceInMsecs(cdates.get(0))).toArray();
-      // make sure we don't try to read from the orgVar again
+      // make sure parent methods do not try to read from the orgVar again
       this.wasRead = true;
     } else {
       super.readValues();
     }
-  }
-
-  @Override
-  public boolean isNumeric() {
-    // we're going to always handle the 1D time coordinate axis case as if it were numeric
-    // because if it is a String or Char, we'll try to convert the values into a
-    // UDUNITS compatible value in the readValues() method.
-    return true;
   }
 
   ////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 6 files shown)


haileyajohnson (commented):
> Reviewed 5 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/568)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @haileyajohnson and @lesserwhirls)

___
*[cdm/core/src/main/java/ucar/unidata/io/RemoteRandomAccessFile.java, line 123 at r1](https://reviewable.io/reviews/unidata/netcdf-java/568#-MMWnOPk73s5bzL7YvJP:-MMWnOPk73s5bzL7YvJQ:bnj30to) ([raw file](https://github.com/unidata/netcdf-java/blob/7a26a8698a9b09474f6c3c2a802c367397429fb5/cdm/core/src/main/java/ucar/unidata/io/RemoteRandomAccessFile.java#L123)):*
> ```Java
>     // to end of the cache block
>     totalBytesRead += readCacheBlockPartial(pos, buff, currentOffsetIntoBuffer, true);
>     currentOffsetIntoBuffer += totalBytesRead;
> ```

What happens if len < readCacheBlockSize and we only want the middle of a block (i.e. don't want to read all the way to the end of the first block)?

lesserwhirls (commented):
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/568)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @haileyajohnson)

___
*[cdm/core/src/main/java/ucar/unidata/io/RemoteRandomAccessFile.java, line 123 at r1](https://reviewable.io/reviews/unidata/netcdf-java/568#-MMWnOPk73s5bzL7YvJP:-MMXONRF5U2Yygdet2Ll:b-uv4b4j) ([raw file](https://github.com/unidata/netcdf-java/blob/7a26a8698a9b09474f6c3c2a802c367397429fb5/cdm/core/src/main/java/ucar/unidata/io/RemoteRandomAccessFile.java#L123)):*

What happens if len < readCacheBlockSize and we only want the middle of a block (i.e. don't want to read all the way to the end of the first block)?
</blockquote></details>

So that case is handled in `readCacheBlockPartial(...)` - we start off assuming that we read to the end of the cache block (line 182), but then we do a few more checks. At line 196 we check to make sure we are not reading past what would be the end of the file (the last cache block will likely not be a full cache block), and then at line 199 we make sure that we are not trying to fit more bytes into the destination byte array than the byte array can actually hold. Those final two checks, specifically that last one, should handle the case that we only want to read a small portion of a single cache block.

That said, I should update the comments in `readFromCache` for both of the partial reads to reflect the following:

For the first block (`forward=true` case), the size that we read is the smaller of the size to the end of the block, the size to the end of the file, or the space we are trying to fill in the destination byte array (full byte array length minus any offset).

For the last block (`forward=false`), we read up to the smaller of either the number of bytes to the end of the file or the size of the destination.

lesserwhirls (commented):
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/568)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @haileyajohnson)

___
*[cdm/core/src/main/java/ucar/unidata/io/RemoteRandomAccessFile.java, line 123 at r1](https://reviewable.io/reviews/unidata/netcdf-java/568#-MMWnOPk73s5bzL7YvJP:-MMXbvkB94CyE5p9HWBs:b-72cien) ([raw file](https://github.com/unidata/netcdf-java/blob/7a26a8698a9b09474f6c3c2a802c367397429fb5/cdm/core/src/main/java/ucar/unidata/io/RemoteRandomAccessFile.java#L123)):*
<details><summary><i>Previously, lesserwhirls (Sean Arms) wrote…</i></summary><blockquote>

So that case is handled in `readCacheBlockPartial(...)` - we start off assuming that we read to the end of the cache block (line 182), but then we do a few more checks. At line 196 we check to make sure we are not reading past what would be the end of the file (the last cache block will likely not be a full cache block), and then at line 199 we make sure that we are not trying to fit more bytes into the destination byte array than the byte array can actually hold. Those final two checks, specifically that last one, should handle the case that we only want to read a small portion of a single cache block.

That said, I should update the comments in `readFromCache` for both of the partial reads to reflect the following:

For the first block (`forward=true` case), the size that we read is the smaller of the size to the end of the block, the size to the end of the file, or the space we are trying to fill in the destination byte array (full byte array length minus any offset).

For the last block (`forward=false`), we read up to the smaller of either the number of bytes to the end of the file or the size of the destination.
</blockquote></details>

Ok, check these changes out and let me know how that sounds: https://github.com/Unidata/netcdf-java/pull/568/commits/5f4b018ee06b67c4b734db9fdbd75bf3b74100f6

haileyajohnson (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/568)* status: all files reviewed, 1 unresolved discussion (waiting on @lesserwhirls)

___
*[cdm/core/src/main/java/ucar/unidata/io/RemoteRandomAccessFile.java, line 123 at r1](https://reviewable.io/reviews/unidata/netcdf-java/568#-MMWnOPk73s5bzL7YvJP:-MMXcmG0D39XsvH0D44G:b-8uzgzb) ([raw file](https://github.com/unidata/netcdf-java/blob/7a26a8698a9b09474f6c3c2a802c367397429fb5/cdm/core/src/main/java/ucar/unidata/io/RemoteRandomAccessFile.java#L123)):*
<details><summary><i>Previously, lesserwhirls (Sean Arms) wrote…</i></summary><blockquote>

Ok, check these changes out and let me know how that sounds: https://github.com/Unidata/netcdf-java/pull/568/commits/5f4b018ee06b67c4b734db9fdbd75bf3b74100f6
</blockquote></details>

:thumbsup: ohp, my bad - I saw read to end of block and thought "wait a minute" and didn't see the rest. looks good!

Issue #572, opened 2020-11-27
valid_range and packed data.


rschmunk:
Working with the latest 5.x maintenance SNAPSHOT.

A user has a test dataset with a variable that looks like
```
    double test_variable(test_dimension=5);
      :valid_range = 0S, 4S; // short
      :scale_factor = -1.0; // double
```

Note the negative scaling factor.

When the dataset is open enhanced, the data are all reported as NaN. The reason for this seems to be that the `isInvalidData` method in `VariableEnhancer` and `EnhanceScaleMissingUnsigned` compares the unpacked data value to the unpacked valid range values. But because the scaling factor is negative, the unpacked minimum is > than the unpacked maximum, and all data values are therefore invalid.

The discussion of valid range attributes and packed data attributes suggests that this is the wrong approach. Specifically, the [Attribute Conventions document](https://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html) states in the description for scale_factor "If valid values are specified using the valid_min, valid_max, valid_range, or _FillValue attributes, those values should be specified in the domain of the data in the file (the packed data), so that they can be interpreted **before the scale_factor and add_offset are applied**."

The user provided an alternative dataset where  the attributes were specified as
```
  :valid_range = 4S, 0S; // short
  :scale_factor = -1.0; // double
```
and in that case the data were evaluated as valid and thus plottable. But it appears extremely odd to someone else glancing at the data description that the attribute specifies a min greater than the max.

If a fix is necessary, it seems the simplest approach would be, after the constructor calculates the unpacked valid min and max values, to compare them and if the min is greater than the max, to swap their values?

lesserwhirls:
Greetings @rschmunk! Excellent breakdown of the issue. I think I have a fix for this along the lines that you suggest and should have something in place tomorrow morning.

ethanrd:
Not sure this affects the underlying issue but the mismatch between data variable and `valid_range` data types caught my eye.

Does the test dataset pass CF compliance checkers? Both CF ([section 8.1](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#packed-data)) and [NUG](https://www.unidata.ucar.edu/software/netcdf/documentation/NUG/best_practices.html#bp_Missing-Data-Values) say that `valid_range` must be the same data type as the packed data. So maybe

```
short test_variable(test_dimension=5);
      :valid_range = 0S, 4S; // short
      :scale_factor = -1.0; // double
```

[...]

PR #575, opened 2020-12-01
Valid min/max for packed data

lesserwhirls:
Properly unpack and set valid min/max when scale is negative.

Fixes Unidata/netcdf-java#572

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/575)


lesserwhirls:
Ah, thanks for catching that! I think `VariableEnhancer` was a class that we thought we would need for the new API to maintain backwards compatibility, but in the end it wasn't needed. Since it was part of a release, we deprecated it and it has been removed in 6. However, it does not look like it's currently used anywhere in our code, although it could be used elsewhere directly, so I'll make the change there as well.

lesserwhirls:
Looks like the Github Actions runs were never able to recover from the issues yesterday, but closing and reopening the PR lined things out. We should be good now.

haileyajohnson (approved):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/575)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lesserwhirls)

PR #579, opened 2020-12-04
Create public_api_feedback.md

lesserwhirls:
Create an issue template for feedback on the public api.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/579)


lesserwhirls:
Also add the default non-security issue template from the Unidata org default templates (will no longer get picked up now that we have an issue template in the repo).

PR #586, opened 2020-12-07
Raf tests

haileyajohnson:
- add tests for RandomAccessFile
- allow endianness to be set on writes
- fix bugs on reading UTF-16 string (readStringMax and readLine)

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/586)


lesserwhirls (changes requested):
Just a few comments:

* Can we note in the main javadoc description for `ucar.unidata.io.RandomAccessFile` that while the intention is to be a drop in replacement for `java.io.RandomAccessFile`, we have made some extensions that are not available in the `java.io` version (and list them)? I think that'd specifically be the addition of methods that let you set the `Charset` on read, and `ByteOrder` on write?
* With the additions above, can we note in the docstrings of those methods that these are not available in `java.io.RandomAccessFile`.

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/586#-:-MO2L1LGDT4o4Z-gayQc:bvjgmhb)* status: all files reviewed, 1 unresolved discussion (waiting on @haileyajohnson)

___
*[cdm/core/src/main/java/ucar/unidata/io/RandomAccessFile.java, line 1641 at r1](https://reviewable.io/reviews/unidata/netcdf-java/586#-MO2KVYH7d9BXSuhHrIJ:-MO2KVYH7d9BXSuhHrIK:b-2mw7l) ([raw file](https://github.com/unidata/netcdf-java/blob/a9347c93b27fe4b970511af8e2c6895c0f0c67dd/cdm/core/src/main/java/ucar/unidata/io/RandomAccessFile.java#L1641)):*
> ```Java
>    * @throws IOException if an I/O error occurs.
>    */
>   public final void writeShort(int v, int endian) throws IOException {
> ```

Do we need the signatures that set endianness using `int`?

haileyajohnson:
___
*[cdm/core/src/main/java/ucar/unidata/io/RandomAccessFile.java, line 1641 at r1](https://reviewable.io/reviews/unidata/netcdf-java/586#-MO2KVYH7d9BXSuhHrIJ:-MO7SA6x3UUP1rEnKQQx:b-stzvhj) ([raw file](https://github.com/unidata/netcdf-java/blob/a9347c93b27fe4b970511af8e2c6895c0f0c67dd/cdm/core/src/main/java/ucar/unidata/io/RandomAccessFile.java#L1641)):*
<details><summary><i>Previously, lesserwhirls (Sean Arms) wrote…</i></summary><blockquote>

Do we need the signatures that set endianness using `int`?
</blockquote></details>

:woman_shrugging: probably not, but we allow specifying endianness as an `int` in the `order` method, so I was trying to be consistent. And it's potentially one less imported package to not use the ByteOrder type? But I'm open to dropping it, that would definitely reduce the number of public methods by a lot.

lesserwhirls (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/586)* status: all files reviewed, 1 unresolved discussion (waiting on @lesserwhirls)

lesserwhirls:
Looks great! Last thing would be a rebase and we should be good to go!

haileyajohnson:
Ooops, meant to do that on the last push - done now =)

lesserwhirls (approved):
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/586)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @haileyajohnson)

lesserwhirls:
Awesome - thanks @haileyajohnson! Once it's ported to `develop`, we can remove the `port: 6.x` label (experimenting with labels to keep tack of what needs ported to where).

PR #593, opened 2020-12-11
PORT: Valid min/max for packed data

### LINKED ISSUE #572: valid_range and packed data.

lesserwhirls:
Port of Unidata/netcdf-java#575 to `develop`. Closes Unidata/netcdf-java#572.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/593)

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/dataset/EnhanceScaleMissingUnsignedImpl.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/EnhanceScaleMissingUnsignedImpl.java
@@ -234,6 +234,14 @@ public class EnhanceScaleMissingUnsignedImpl implements EnhanceScaleMissingUnsig
             validMax = applyScaleOffset(validMax);
           }
         }
+        // During the scaling process, it is possible that the valid minimum and maximum values have effectively been
+        // swapped (for example, when the scale value is negative). Go ahead and check to make sure the valid min is
+        // actually less than the valid max, and if not, fix it. See https://github.com/Unidata/netcdf-java/issues/572.
+        if (validMin > validMax) {
+          double tmp = validMin;
+          validMin = validMax;
+          validMax = tmp;
+        }
       }
     }
   }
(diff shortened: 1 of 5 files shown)


haileyajohnson (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/593)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lesserwhirls)

PR #611, opened 2021-02-22
Fix issue #103, ternary operator result must have the same type.

### LINKED ISSUE #103: NcMLWriter does not write integer variable values as integers
In NetCDF-Java version 5.0.0:

In method `makeValuesElement(Variable, boolean)` of class `NcMLWriter` there obviously is the intention to separate writing of floating point values and integer values (by the boolean variable `isRealType`), but appending to the string builder (buffer) is always made through `StringBuilder.append(double)` since the used ternary operator will always have a result of type `double`. A statement like `if (isRealType) buff.append(iter.getDoubleNext()); else buff.append(iter.getIntNext());` should be used instead.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/611)


lesserwhirls:
Review in progress. Includes Unidata/netcdf-java#608, Unidata/netcdf-java#609, and Unidata/netcdf-java#610.

lesserwhirls (approved):
> Reviewed 127 of 127 files at r1, 84 of 84 files at r2, 28 of 28 files at r3, 217 of 217 files at r4, 13 of 13 files at r5, 2 of 2 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/611)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnLCaron)

___
*[cdm/core/src/main/java/ucar/unidata/io/http/HTTPRandomAccessFile.java, line 154 at r4](https://reviewable.io/reviews/unidata/netcdf-java/611#-MUJuH869klazL_Z1OIp:-MUJuH869klazL_Z1OIq:bnkpmml) ([raw file](https://github.com/unidata/netcdf-java/blob/88fe3727477c775ef5c602d3bae9b5072f6cf678/cdm/core/src/main/java/ucar/unidata/io/http/HTTPRandomAccessFile.java#L154)):*
> ```Java
>     System.out.println(title);
>     for (Map.Entry<String, String> entry : headers) {
>       System.out.printf("  %s = %s" + entry.getKey() + "%n", entry.getValue());
> ```

Minor: Is there an extra `%s` format flag in this call? Should it be:

`System.out.printf("  %s = " + entry.getKey() + "%n", entry.getValue())`

or

`System.out.printf("  %s = %s%n", entry.getKey(), entry.getValue())`

If so, we can pick this change up in another PR.

lesserwhirls:
Thank you @JohnLCaron! I'll go ahead and merge this in (and as I said in the comment, if that `printf` line needs changed, it can be added in another PR).

PR #614, opened 2021-02-24
For Experimental HRRR GRIB2 tables, use NCLName

lesserwhirls:
Previously we were using the `WGrib2Name` from the HRRR Grib2 table, which is not even close to being unique (multiple
discipline, category, parameter combinations share the same `WGrib2Name` in their table). The `NCLName` isn't unique either,
but at least we don't see collisions in the data we currently receive.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/614)


haileyajohnson (approved):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/614)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lesserwhirls)

PR #617, opened 2021-02-26
 Fix for Issue #606, deal with interval time units different than pds. 

### LINKED ISSUE #606: Bug in NDFD time intervals when interval units different from pds units
Ticket ID: JMQ-641023

Some NDFD datasets use time intervals that are different from PDS time unit. This is not being handled correctly in version 5 and 6. Haven't tested 4.6, but maybe broken there also.

Im working on a fix for 6, probably can be easily backported to 5.

In looking at this, Im reconsidering how time intervals are calculated, and its possible they are still wrong, at least for NDFD.
Spot checking GFS/global 0.25, NAM conus 12, NAM firewxnest, RR conus 13, CMC/RDPS/NA_15km appear ok. NOAA_GSD/HRRR/CONUS_3km/surface has 88/122261 possibly wrong. 

1. How to get the truth? Im guessing we need to fire up the "standard" NCEP library and compare. Its possible that different groups handle it differently, since I was following someone's advice. Or that the only bug is when the units change.
2. Check to see what 4.6 is doing.
3. Fix 6 and then 5 once we know what its supposed to be.
4. What about ECMWF and others?

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/617)

### DIFF
--- a/cdm/core/src/main/java/thredds/filesystem/ControllerOS.java
+++ b/cdm/core/src/main/java/thredds/filesystem/ControllerOS.java
@@ -103,7 +103,7 @@ public class ControllerOS implements MController {
 
   ////////////////////////////////////////////////////////////
 
-  // handles filtering and removing/including subdirectories
+  // handles filtering and removing/including subdirectories LOOK use AbstractIterator
   private static class FilteredIterator implements Iterator<MFile> {
     private final Iterator<MFile> orgIter;
     private final CollectionConfig mc;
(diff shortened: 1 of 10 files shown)


JohnLCaron:
Ready, and fixes issue #606

lesserwhirls (approved):
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/617)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnLCaron)

lesserwhirls:
Awesome - thanks @JohnLCaron!

PR #626, opened 2021-03-04
Convert DataType into ArrayType, deprecations where needed in public API.

JohnLCaron:
Trying to get the public API, many more internal ones to do.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/626)

PR #630, opened 2021-03-16
Public API tweaks.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/630)


JohnLCaron:
Looks like httpservices test is failing. 

Not sure why this is happening now. Did these tests just get added or reenabled?

lesserwhirls:
We are relying on a SNAPSHOT version of some TDS artifacts to do some basic testing. I had made a change on the tds side, which passed all of the github action tests, but ended up breaking some authentication related things (I think...spring-security upgrade). Definitely points to a critical hole in the tds tests that run on github actions, but also an unnecessarily complicated test relationship that still exists between the two project. In the meantime, we can revert the versions of those war files to `beta8` from `SNAPSHOT` and the tests will work again (see my comment over at https://github.com/Unidata/netcdf-java/pull/626#issuecomment-801147489)

JohnLCaron:
Ok, thanks. I made the change in the next PR.

PR #634, opened 2021-03-22
Fix cmdr test failures.

JohnLCaron:
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/634)

### DIFF
--- a/cdm/core/src/main/java/ucar/array/Array.java
+++ b/cdm/core/src/main/java/ucar/array/Array.java
@@ -72,8 +72,7 @@ public abstract class Array<T> implements Iterable<T> {
     return indexFn.length();
   }
 
-  @Override
-  public String toString() {
+  public String show() {
     StringBuilder sbuff = new StringBuilder();
     boolean first = true;
     for (T value : this) {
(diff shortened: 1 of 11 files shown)


lesserwhirls:
Can you merge in the latest changes?

lesserwhirls:
...dang it...it used to be that there was only one object under the AWS S3 bucket root, so to say, but now there are two (they added a user guide), which broke a poor assumption for my test. I will need to do something slightly different. For now, I think we can add `@Category(NeedsExternalResource.class)` for `cdm/s3/src/test/java/thredds/filesystem/s3/TestControllerS3.java`, which will at least let us move along with your PR.

lesserwhirls (approved):
Thank you @johnlcaron, and sorry for the extra hoops.

> Reviewed 10 of 10 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/634#-:-MWVwUufAryXp4WTWoeL:baxpxf3)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnLCaron)

PR #650, opened 2021-04-13
Fix jdk14 logging.

JohnLCaron:
Remove "Replace every instance of "commons-logging" in the dependency tree with "jcl-over-slf4j" in dependencies.gradle. This interferes with ability to control grpc logging.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/650)


JohnLCaron:
Looks like ":cdm-lite:AppWarBeforeIntegrationTest" is failing??

JohnLCaron:
I wonder if we should get rid of gretty?

lesserwhirls:
Pulled in with Unidata/netcdf-java#660. Closing.

PR #682, opened 2021-05-13
Zarr support infrastructure

haileyajohnson:
This pull request has the infrastructure for zarr support, i.e. distributed randomaccessfile reads.
- cdm-zarr subproject is created
- iosp is stubbed out but not implemented in this PR to keep it smaller
- RandomAccessDirectory is implemented
- RandomAccessDirectoryItem interface and VirtualRandomAccesFile are implemented to allow for lazy loads and save space, particularly for zip files (items are not decompressed until a read attempt is made)
- MFileZip and MController are implemented to represent Zip files and individual Zip entries
- NetcdfFiles now supports decompressing a single zip entry within a zip file
- addition of test files (37 in a directory + 1 zip file) representing different data types (int, double, string, etc.)

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/682)


cdm/zarr/src/main/java/ucar/unidata/io/zarr/RandomAccessDirectory.java
```diff
@@ -0,0 +1,270 @@
...
+    this(location, RandomAccessFile.defaultBufferSize);
+  }
+
+  public RandomAccessDirectory(String location, int bufferSize) throws IOException {
+    super(bufferSize);
+    this.bufferSize = bufferSize;
+    this.location = location.replace("\\", "/"); // standardize path
+    this.readonly = true; // RandomAccessDirectory does not support writes
+
+    // build children list
```
haileyajohnson, line 50: I ended up flattening the tree, but I'm not sure if it was the right move (it will probably become clear one way or the other as I work on the IOSP), and I might bring back the tree.... the tradeoff is basically that a lot of reads from the same file are faster with the flattened structure, but searching is faster with the tree. So it really depends on whether we'll be accessing different files or the same file more often.

lesserwhirls (changes requested):
Just a few minor things, but overall looks really good - pretty intense chunk of code down in the underbelly of the library!

> Reviewed 57 of 57 files at r1.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/682#-:-M_cBQM9-2c10De_CRf4:bf8ibnw)* status: all files reviewed, 6 unresolved discussions (waiting on @haileyajohnson)

___
*[cdm/core/src/main/java/ucar/nc2/NetcdfFiles.java, line 395 at r1](https://reviewable.io/reviews/unidata/netcdf-java/682#-M_buW-Z1-8e1MCVrUlM:-M_buW-Z1-8e1MCVrUlN:b-fgita7) ([raw file](https://github.com/unidata/netcdf-java/blob/681645b93c66bc1969795bca73362c3aae212d5e/cdm/core/src/main/java/ucar/nc2/NetcdfFiles.java#L395)):*
> ```Java
>   }
> 
>   public static ucar.unidata.io.RandomAccessFile getRaf(String location, int buffer_size) throws IOException {
> ```

Since this will effectively be adding `getRaf` to the public API, we'll need to add javadocs.
___
*[cdm/core/src/main/java/ucar/unidata/io/RandomAccessFile.java, line 35 at r1](https://reviewable.io/reviews/unidata/netcdf-java/682#-M_cCSxb7F83B1rzJA13:-M_cCSxb7F83B1rzJA14:bf6aybz) ([raw file](https://github.com/unidata/netcdf-java/blob/681645b93c66bc1969795bca73362c3aae212d5e/cdm/core/src/main/java/ucar/unidata/io/RandomAccessFile.java#L35)):*
> ```Java
>  * <li>Read String methods support user-specified Charsets (default UTF-8).</li>
>  * <li>Support for both big and little endiannness on reads and write: users may specify the byte order for I/O
>  * operations.</li>
> ```

Maybe add something here about how there has been an extension to support the concept of a Random Access Directory?
___
*[cdm/core/src/main/java/ucar/unidata/io/RandomAccessFile.java, line 471 at r1](https://reviewable.io/reviews/unidata/netcdf-java/682#-M_byj9PCcOz4RwJfv1o:-M_byj9PCcOz4RwJfv1p:b72kyap) ([raw file](https://github.com/unidata/netcdf-java/blob/681645b93c66bc1969795bca73362c3aae212d5e/cdm/core/src/main/java/ucar/unidata/io/RandomAccessFile.java#L471)):*
> ```Java
>   }
> 
>   /**
> ```

I'm not always good at this, but trying to get better at following [this style guide](https://google.github.io/styleguide/javaguide.html#s7.3.1-javadoc-exception-self-explanatory).

Could reduce this javadoc to:

~~~java
/** Returns true if RandomAccessFile is a directory structure. */
~~~

or maybe

~~~java
/** Returns true if RandomAccessFile represents a directory structure. */
~~~
___
*[cdm/zarr/build.gradle, line 15 at r1](https://reviewable.io/reviews/unidata/netcdf-java/682#-M_c1Y9GCIVHO-RcBx6z:-M_c1Y9GCIVHO-RcBx7-:b-lljjtb) ([raw file](https://github.com/unidata/netcdf-java/blob/681645b93c66bc1969795bca73362c3aae212d5e/cdm/zarr/build.gradle#L15)):*
> ```Gradle
>     api project(':cdm:cdm-s3')
> 
>     compile 'org.jdom:jdom2'
> ```

I think all of these `compile` dependencies can become `implementation` dependencies...I think...it will either compile or not.
___
*[cdm/zarr/src/main/java/thredds/filesystem/zarr/ControllerZip.java, line 28 at r1](https://reviewable.io/reviews/unidata/netcdf-java/682#-M_c1Dm41r24_fBsEIHw:-M_c1Dm41r24_fBsEIHx:b2lr3ny) ([raw file](https://github.com/unidata/netcdf-java/blob/681645b93c66bc1969795bca73362c3aae212d5e/cdm/zarr/src/main/java/thredds/filesystem/zarr/ControllerZip.java#L28)):*
> ```Java
> public class ControllerZip extends ControllerOS implements MController {
> 
>   private static org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ControllerZip.class);
> ```

You can make this logger `final`.
___
*[cdm/zarr/src/main/java/ucar/unidata/io/zarr/RandomAccessDirectory.java, line 50 at r1](https://reviewable.io/reviews/unidata/netcdf-java/682#-M_bTHsO3P4--mLhhmSZ-r1-50:-M_c26ER5h6fpGg4tP2P:b61gg39) ([raw file](https://github.com/unidata/netcdf-java/blob/681645b93c66bc1969795bca73362c3aae212d5e/cdm/zarr/src/main/java/ucar/unidata/io/zarr/RandomAccessDirectory.java#L50)):*

I ended up flattening the tree, but I'm not sure if it was the right move (it will probably become clear one way or the other as I work on the IOSP), and I might bring back the tree.... the tradeoff is basically that a lot of reads from the same file are faster with the flattened structure, but searching is faster with the tree. So it really depends on whether we'll be accessing different files or the same file more often.
</blockquote></details>

Sounds like a good plan to me. I could see it coming down to access patterns and chunk sizes, but that might shake out in testing.

lesserwhirls (approved):
:lgtm_strong: Thanks for all the hard work @haileyajohnson! w00t!

> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/682#-:-M_fzl9A2oNJ7RSpeo-j:bq6wksy)* status: all files reviewed, 1 unresolved discussion (waiting on @haileyajohnson)

PR #696, opened 2021-06-07
Use java.net.http instead of ucar.httpservices.

JohnLCaron:
Move http processing to internal.
Remove support for DAP4.
Not yet managing user credentials / authentication.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/696)


JohnLCaron:
The opendap module tests depend on a local dods server:

2021-06-07T13:56:15.9768103Z     java.io.IOException: http://localhost:8080/dts/test.50.dods?types[0:1:9] is not a valid URL, return status=400

Which presumably I have clobbered. Perhaps something to do with farmBeforeIntegrationTest target? 

Sorry, Ive lost track of how that works.

JohnLCaron:
Maybe thats what cdm-lite is for? Which I didnt (yet) muck with. Which brings up the issue of how to replace that functionality, since those are synthetic datasets. Hmmmm.

lesserwhirls:
So things are working, but the main issue is that the query isn't encoded:

~~~
HTTP Status 400 – Bad Request
Type Exception Report

Message Invalid character found in the request target. The valid characters are defined in RFC 7230 and RFC 3986

Description The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

Exception

java.lang.IllegalArgumentException: Invalid character found in the request target. The valid characters are defined in RFC 7230 and RFC 3986
	org.apache.coyote.http11.Http11InputBuffer.parseRequestLine(Http11InputBuffer.java:483)
	org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:502)
	org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:65)
	org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:810)
	org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1506)
	org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49)
	java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
	java.base/java.lang.Thread.run(Thread.java:834)
Note The full stack trace of the root cause is available in the server logs.
~~~

With this specific request, it's the square brackets that cause the problem. Does the new java api include a solid url encoder (one that handles the different parts of a URI properly, like paths vs query)?

JohnLCaron:
Ok, thanks for finding that. I did munge the escaping. Ill get a fix.

And, is that local server coming from cdm-lite? If so, how to replace?

lesserwhirls:
It turns out that the local server is coming from the war artifact of `edu.ucar:dtswar:5.0.0-beta8` - I hadn't moved the `opendap` subproject to use `cdm-lite` for its test server.

lesserwhirls:
Also, not sure if you saw this or not, but in netCDF-Java `4.6.x`, `5.x`, and `6.x`, the choice to encode URLs or not is controlled by a Java System Property `httpservices.urlencode` (default: `true`). This is because there are quite a few servers and front-end proxies that don't handle encoded URLs properly, and it's, unfortunately, useful and necessary to be able to turn it off.

JohnLCaron:
Well, I wrote a test (TestDConnectWithTestServer) which uses thredds-test to make some opendap data requests with [...] in the query string. It worked correctly whether or not I escaped the query.

```
/*
 * DConnect2.openConnection
 * https://thredds-test.unidata.ucar.edu/thredds/dodsC/grib/NCEP/GFS/CONUS_80km/Best.dods?time%5B0:42%5D
 * Dataset {
 * Float64 time[time = 43];
 * } grib/NCEP/GFS/CONUS_80km/Best;
 * Float64 time[time = 43] = {0.0, 6.0, 12.0, 18.0, 24.0, 30.0, 36.0, 42.0, 48.0, 54.0, 60.0, 66.0, 72.0, 78.0, 84.0,
 * 90.0, 96.0, 102.0, 108.0, 114.0, 120.0, 126.0, 132.0, 138.0, 144.0, 150.0, 156.0, 162.0, 168.0, 174.0, 180.0, 186.0,
 * 192.0, 198.0, 204.0, 210.0, 216.0, 222.0, 228.0, 234.0, 240.0, 246.0, 252.0};
 * 
 * 
 * DConnect2.openConnection
 * https://thredds-test.unidata.ucar.edu/thredds/dodsC/grib/NCEP/GFS/CONUS_80km/Best.dods?time[0:42]
 * Dataset {
 * Float64 time[time = 43];
 * } grib/NCEP/GFS/CONUS_80km/Best;
 * Float64 time[time = 43] = {0.0, 6.0, 12.0, 18.0, 24.0, 30.0, 36.0, 42.0, 48.0, 54.0, 60.0, 66.0, 72.0, 78.0, 84.0,
 * 90.0, 96.0, 102.0, 108.0, 114.0, 120.0, 126.0, 132.0, 138.0, 144.0, 150.0, 156.0, 162.0, 168.0, 174.0, 180.0, 186.0,
 * 192.0, 198.0, 204.0, 210.0, 216.0, 222.0, 228.0, 234.0, 240.0, 246.0, 252.0};
 */
```

But the git test uses dtswar, which fails. So what is dtswar?

JohnLCaron:
> This is because there are quite a few servers and front-end proxies that don't handle encoded URLs properly, and it's, unfortunately, useful and necessary to be able to turn it off.

Do you have a concrete example of a server where you had to use that flag? Is it random, or all servers built on some technology, or ?

The problem with httpservices.urlencode is that its a global flag.

lesserwhirls:
> Well, I wrote a test (TestDConnectWithTestServer) which uses thredds-test to make some opendap data requests with [...] in the query string. It worked correctly whether or not I escaped the query.

Our TDSs are behind an Apache proxy, which seems to be quite happy to let the unencoded brackets through. However, by default, Tomcat will reject unencoded URLs with unsafe characters, which is why the tests fail when running against gretty (which does not have the apache frontend).

> But the git test uses dtswar, which fails. So what is dtswar?

`dtswar` the "dap test server", which is the war file that is used to run the dap test server used by netCDF-C (built as part of the TDS build). It's like cdm-lite...sort of, but not quite. There is a [publicly accessible server running it](https://remotetest.unidata.ucar.edu/dts/), but I'm not certain how stable it is or what version it is using.

The best way I could figure out how to deal with the encoding was a system property at runtime, which isn't ideal. It's not static, so it can change at runtime, but that does not make it better. Tomcat will throw a 400, so we could retry on the 400 response, but other servers like the one at Columbia do not throw a 400. For example, this works:

http://iridl.ldeo.columbia.edu/SOURCES/.NOAA/.NCEP/.CPC/.GLOBAL/.daily/dods.dods?olr[0:1:0][0:1:72][0:1:143]

but this returns a 404:

http://iridl.ldeo.columbia.edu/SOURCES/.NOAA/.NCEP/.CPC/.GLOBAL/.daily/dods.dods?olr%5B0:1:0%5D%5B0:1:72%5D%5B0:1:143%5D

I've also seen NOAA servers screw up when doing "value added" redirects from `http` -> `https` via apache, in that they not only redirect but also try to percent encode the redirected url before passing it to the TDS backend...which results in a double percent encoding if the original URL is already encoded. I reached out to a few of those sites directly, but only one fixed the issue by not trying to handle encoding on the server side. They probably apply the encoding to the redirect because of older clients that are abandoned and will likely not be updated to encode before making the request.

JohnLCaron:
Well, that sounds like quite the mess. it seems that encoding is the right thing to do from a standards POV, and has to be the default. Do you think this problem is mostly with old DODS servers, or ?

JohnLCaron:
Ok, I found the dts artifact in https://github.com/Unidata/tds.git and have it checked out.

How does one build and start up dtswar?

lesserwhirls:
There currently isn't a way to run the dtswar from the TDS repo (only the dap4 test server (d4ts) and the tds). I could add that in though. To build, just run:

~~~
./gradlew :opendap:dtwsar:war
~~~

From the netCDF-java `develop` branch, you can run the dts using:

~~~
./gradlew :opendap:farmRun
~~~

which should result in the following output in the terminal:

~~~
11:14:41 INFO  Tomcat 8.5.49 started and listening on ports 8080, 8443
11:14:41 INFO  dts runs at:
11:14:41 INFO    http://localhost:8080/dts
11:14:41 INFO    https://localhost:8443/dts

> Task :opendap:farmRun
Press any key to stop the server.
~~~

As it says, you can shutdown the server by pushing any key in the terminal. This is similar to what happens when you run `./gradlew :opendap:test`, except that task uses the `farmBeforeIntegrationTest` and `farmAfterIntegrationTest` tasks to allow gradle to handle startup and shutdown of the server before and after the tests run.

If you want to debug things on the server side (step though code at least), you can use the `./gradlew :opendap:farmRunDebug` task and attach a debugger to the port it opens.

gretty fetches and uses a pre-built dts war file from the artifacts server, so there isn't anything to build when running from netCDF-Java.

JohnLCaron:
Added issue Unidata/netcdf-java#697 and Unidata/netcdf-java-ng#1 about Opendap.

JohnLCaron:
ready

lesserwhirls (approved):
Looks good to me! The new `java.net.http` code seems pretty nice. Do you think we'll be able to remove the `httpservices` gradle subproject all together in 7?

> Reviewed 37 of 38 files at r1, 2 of 2 files at r2, 2 of 4 files at r3, 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/696#-:-MbhA16v8SgFknC3b6wJ:bueacc3)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnLCaron)

JohnLCaron:
Yes, we will remove httpservices. Good to have in for now as we can compare the two.

PR #699, opened 2021-06-08
Disentangle ncstream, in preparation for deleting.

JohnLCaron:
Cant delete yet, because its used in dt, ft, ft2, which is not yet ready.
This removes its use everywhere else.
Also remove NetcdfFileFormat.NCSTREAM.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/699)


JohnLCaron:
ready

PR #711, opened 2021-06-11
Restore Grib CoverageCollection, needed for testing.

JohnLCaron:
There is lots of work to be done on nc2.grid, and I need ft2.coverage for comparison and testing. So Im restoring that for now. 

This will fix lots of Jenkins failures. Not sure why its hanging, should have just failed tests I think?

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/711)


JohnLCaron:
ready

PR #849, opened 2021-09-24
remove httpservices test code

JohnLCaron:
It looks like the main code was already removed, and this is some dead test code.

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/849)


JohnLCaron:
So this is failing in the same place. Not sure why. Some weirdness in UnitTestCommon.locateThreddsRoot().

PR #873, opened 2021-10-08
Time coordinate values must use Longs, not Integers.

JohnLCaron:
I think this happened long ago when we were switching to CalendarDate?
But only triggers when regenerating grib ncx4.
So why didnt we see more Jenkins failures??

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/873)


JohnLCaron:
ready. can you start a new jenkins run when this is in? thanks!

PR #916, opened 2021-10-28
[5.x]: Modify XML Schema and DTD URLs for Unidata hostname switch

ethanrd:
## Description of Changes

Unidata hosted XML Schema and DTD files are now available from "https://schemas.unidata.ucar.edu/*" (with redirects in place from the old URLs, "http://www.unidata.ucar.edu/schemas/*"). This change updates all occurrences.

## PR Checklist
       until ready for review


ethanrd:
CatalogGenConfig.*.xsd is referenced in some cdm-test files and in legacy code but is itself no longer in resources/thredds/schema directory. Could consider adding it to legacy/src/main/resources/resources/thredds/schemas/ but probably not needed.

The legacy module might be a better place for queryCapability.0.4.xsd as well. But would have to make sure those resources make it into ncIDV.jar. Probably not worth effort at the moment.

I will get this PR ready for review shortly.

haileyajohnson (approved):

PR #917, opened 2021-10-29
Fix intellij code coverage.

JohnLCaron:
Change junit version.
Interfering with IntelliJ code coverage: "Due to a bug in JUnit, junit-vintage-engine:5.6.2 did not support JUnit 4.13.2. It only supported 4.12 or 4.13". See https://stackoverflow.com/questions/67563862/testengine-with-id-junit-vintage-failed-to-discover-tests-junitexception-fa

---
This change is [Reviewable](https://reviewable.io/reviews/unidata/netcdf-java/917)


JohnLCaron:
ready

PR #920, opened 2021-11-01
[5.x]: Clean up Unidata XML Namespace URIs mistakenly converted to "https:"

ethanrd:
## Description of Changes

Some code and document uses of Unidata XML Namespace URIs were converted from “http:” to “https:” when Unidata servers started requiring HTTPS. This change reverts all uses of Unidata XML Namespace URIs back to “http:” form.

## PR Checklist
       until ready for review

### DIFF
--- a/cdm/core/src/main/java/thredds/client/catalog/Catalog.java
+++ b/cdm/core/src/main/java/thredds/client/catalog/Catalog.java
@@ -30,6 +30,11 @@ public class Catalog extends DatasetNode {
   public static final String CATALOG_NAMESPACE_10 = "http://www.unidata.ucar.edu/namespaces/thredds/InvCatalog/v1.0";
   public static final Namespace defNS = Namespace.getNamespace(CATALOG_NAMESPACE_10);
   public static final String NJ22_NAMESPACE = "http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2";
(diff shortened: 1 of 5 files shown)


ethanrd:
There is code in thredds.client.catalog.Catalog to allow for both the "http:" and "https:" forms of the NcML Namespace URI (also a test in TestNcmlRead using testReadHttps.xml). That code and test files could be removed at some point. Though perhaps it is useful if there is much use of the "https:" form in the wild.

haileyajohnson:
@ethanrd is this ready?

ethanrd:
Hi @haileyajohnson - Yes, I believe the changes here are ready. Sorry, I meant to ask what you thought about getting rid of the code that handles occurrences of the NcML "https:" namespace URI. Maybe we should just accept this PR and move thinking about the "https:" code to another issue?

On a slight side note, do you know if the version 8 branch  needs the XML schema and namespace changes?

dopplershift:
@ethanrd This PR can't be merged until you mark it "Ready for review" since it's currently a draft.

ethanrd:
Hi @dopplershift - Yes. I was perhaps being overly cautious with my questions before marking ready. But question is really about doing more from here rather than about not doing this change at all. So, right you are. I will mark as ready for review.

haileyajohnson (approved):

PR #966, opened 2021-12-30
[5.x] work on filters package

haileyajohnson:
## Description of Changes

- changes to the filters api
- share filters between hdf5 and zarr
- rewrite docs to use new api

## PR Checklist

       until ready for review


cdm/core/src/main/java/ucar/nc2/filter/Checksum32.java
```diff
@@ -19,7 +19,16 @@
 public class Checksum32 extends Filter {
 
   public enum CType {
-    FLETCHER, ADLER, CRC
+    FLETCHER("fletcher32", 3), ADLER("adler32", -1), CRC("crc32", -1);
```
tdrwenski, line 22: Should the second and third types have distinct ids?

cdm/core/src/main/java/ucar/nc2/filter/Filters.java
```diff
@@ -10,38 +10,95 @@
...
-   * Create a filter object for the given property list
+   * Create a filter object matching the provided id
    * 
    * @param properties
-   * @return A filter object with the provided properties
+   * @return A filter object with the given properties
    * @throws UnknownFilterException
    */
-  public static Filter getFilterByName(Map<String, Object> properties) throws UnknownFilterException {
+  public static Filter getFilter(Map<String, Object> properties) throws UnknownFilterException {
```
tdrwenski, line 49: is it OK to change a public method name here? We could also have `getFilterByName` and `getFilterById` and then `getFilter` could call those.

cdm/core/src/main/java/ucar/nc2/filter/Filters.java
```diff
@@ -10,38 +10,95 @@
...
     if (properties == null) {
       return nullFilter;
     }
-    String name = (String) properties.get("id");
-    if (name == null || name.isEmpty()) {
+    // read id properties
+    String name = (String) properties.get(Keys.NAME);
+    Object oid = properties.get(Keys.ID);
+    // if no id or name, return null filter
+    if ((name == null || name.isEmpty()) && (oid == null || !(oid instanceof Number))) {
```
tdrwenski: second condition could be simplified
```suggestion
    if ((name == null || name.isEmpty()) && !(oid instanceof Number)) {
```

cdm/core/src/main/java/ucar/nc2/filter/Filters.java
```diff
@@ -10,38 +10,95 @@
...
+      // look for dynamically loaded filters by name
+      for (FilterProvider fp : ServiceLoader.load(FilterProvider.class)) {
+        if (fp.canProvide(name)) {
+          return fp.create(properties);
+        }
+      }
+    }
+
+    // try by id next
+    int id = ((Short) oid).intValue();
```
tdrwenski, line 72: maybe we should check that `oid instanceof Short` before casting here?

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5tiledLayoutBB.java
```diff
@@ -56,25 +60,36 @@ public class H5tiledLayoutBB implements LayoutBB {
...
   /**
    * Constructor.
-   * This is for HDF5 chunked data storage. The data is read by chunk, for efficency.
+   * This is for HDF5 chunked data storage. The data is read by chunk, for efficiency.
    *
    * @param v2 Variable to index over; assumes that vinfo is the data object
    * @param wantSection the wanted section of data, contains a List of Range objects. must be complete
    * @param raf the RandomAccessFile
-   * @param filters set of filters that have been applied to the data
+   * @param filterProps set of filter propertoes from which filter object will be created
```
tdrwenski: typo 
```suggestion
   * @param filterProps set of filter properties from which filter object will be created
```

cdm/core/src/main/java/ucar/nc2/filter/Checksum32.java
haileyajohnson: Fletcher is the only one that has an ID in HDF5, so the others should only be using a name

cdm/core/src/main/java/ucar/nc2/filter/Filters.java
haileyajohnson: I think it's fine because no one was using this package, it was still experimental

cdm/core/src/main/java/ucar/nc2/filter/Filters.java
haileyajohnson: we already checked that it's a Number, I think any Number type can cast to a Short without an exception

tdrwenski: yeah, maybe I am overthinking it here, but could it happen that a filter does have a `name` but no provider for that name? then you reach this code without having checked if the `oid instanceof Number`

tdrwenski (approved):

PR #988, opened 2022-03-17
[5.x] Udt 134933 s3 fileserver problems

tdrwenski:
-Add writeToStream method for MFile
-Add writeToStream method for a byte range of MFile
-Add tests

### DIFF
--- a/cdm/core/src/main/java/thredds/inventory/CollectionManagerCatalog.java
+++ b/cdm/core/src/main/java/thredds/inventory/CollectionManagerCatalog.java
@@ -152,6 +152,11 @@ public class CollectionManagerCatalog extends CollectionManagerAbstract implemen
     public void writeToStream(OutputStream outputStream) throws IOException {
       throw new IOException("Writing MFileRemote not implemented. Filename: " + getName());
     }
+
+    @Override
+    public void writeToStream(OutputStream outputStream, long offset, long maxBytes) throws IOException {
+      throw new IOException("Writing MFileRemote not implemented. Filename: " + getName());
+    }
   }
 
   ///////////////////////////////
--- a/cdm/core/src/main/java/thredds/inventory/CollectionManagerCatalog.java
+++ b/cdm/core/src/main/java/thredds/inventory/CollectionManagerCatalog.java
@@ -5,6 +5,7 @@
 
 package thredds.inventory;
 
+import java.io.OutputStream;
 import thredds.client.catalog.Access;
 import thredds.client.catalog.Dataset;
 import thredds.client.catalog.tools.CatalogCrawler;
@@ -145,6 +146,11 @@ public class CollectionManagerCatalog extends CollectionManagerAbstract implemen
     @Override
     public void setAuxInfo(Object info) {
       this.info = info;
+    }
+
+    @Override
+    public void writeToStream(OutputStream outputStream) throws IOException {
+      throw new IOException("Writing MFileRemote not implemented. Filename: " + getName());
     }
   }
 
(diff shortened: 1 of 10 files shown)


cdm/zarr/src/test/java/thredds/inventory/zarr/TestMFileZip.java
```diff
@@ -0,0 +1,81 @@
+package thredds.inventory.zarr;
```
tdrwenski, line 1: I just realized the `MFileZip:MFileProvider` doesn't get loaded by TDS (so TDS can only have `MFileS3` or `MFileOS` implementations returned from `MFiles.create`). So I'm not sure if these tests are very useful/important-- maybe I should remove them.

cdm/core/src/test/java/thredds/filesystem/TestMFileOS.java
```diff
@@ -0,0 +1,72 @@
...
+    final int endPosition = Math.min(offset + maxBytes, expectedSize);
+
+    assertThat(outputStream.size()).isEqualTo(Math.max(0, endPosition - startPosition));
+
+    final byte[] partialFile = Arrays.copyOfRange(Files.readAllBytes(file.toPath()), startPosition, endPosition);
+    assertThat(outputStream.toByteArray()).isEqualTo(partialFile);
+  }
+
+  private File createTemporaryFile(int size) throws IOException {
+    final File tempFile = File.createTempFile("TestMFileOS-", ".tmp");
```
haileyajohnson: Would you mind changing this to use junit,rule.TemporaryFolder? Any files created in that get cleaned up automatically at the end of the test method, e.g.
```
  @ClassRule
  public static final TemporaryFolder tempFolder = new TemporaryFolder();
  File file = tempFolder.newFile();
```

cdm/core/src/main/java/thredds/inventory/CollectionManagerCatalog.java
```diff
@@ -146,6 +147,16 @@ public Object getAuxInfo() {
...
+    @Override
+    public void writeToStream(OutputStream outputStream) throws IOException {
+      throw new IOException("Writing MFileRemote not implemented. Filename: " + getName());
+    }
+
+    @Override
+    public void writeToStream(OutputStream outputStream, long offset, long maxBytes) throws IOException {
+      throw new IOException("Writing MFileRemote not implemented. Filename: " + getName());
+    }
   }
```
haileyajohnson, line 160: I think this has the potential to break things down to line, but I also think that's a "down the line" problem that we can ignore for now.

cdm/zarr/src/test/java/thredds/inventory/zarr/TestMFileZip.java
haileyajohnson: I think you should leave them, we're going to have to add MFileZip to the TDS if we want to support Zarr zip stores, which we probably do.

haileyajohnson (changes requested):

cdm/core/src/main/java/thredds/inventory/CollectionManagerCatalog.java
tdrwenski: I could also implement these, it is probably not very hard. But not necessary for this bug fix so I could do it in a separate PR

haileyajohnson: Yeah, up to you, but I think it's fine to wait and see how this first pass goes over in the TDS

cdm/core/src/test/java/thredds/filesystem/TestMFileOS.java
tdrwenski: Fixed in 93080b86dc

haileyajohnson (approved):

PR #1004, opened 2022-04-27
remove custom trust strategy

haileyajohnson:
- replace custom trust strategy with default
- general clean up


haileyajohnson:
separating these two changes, didn't mean to wrap them up together

Issue #980, opened 2022-02-10
[VERSION]: 5.x, 6.x, 7.x Alteration of time intervals by 5s


XSalard:
### Versions impacted by the bug

v5.x, v6.x, v7.x

### What went wrong?

In the [CalendarPeriod.getOffset](https://github.com/Unidata/netcdf-java/blob/b24becaf96e45e1cf21c8f45c5e168f12dddd289/cdm/core/src/main/java/ucar/nc2/time/CalendarPeriod.java#L247) function, data is usually altered as to add a supplementary 5s to the time difference. 
Is there any reason for modifying the data in such a way with this "5 second slop"? 
I don't believe i can share my data, but this does alter it upon reading.

### If you have an example file that you can share, please attach it to this issue.
If so, may we include it in our test datasets to help ensure the bug does not return once fixed?
Note: the test datasets are publicly accessible without restriction.

No

### Code of Conduct

- [X] I agree to follow the UCAR/Unidata Code of Conduct

rschmunk:
This may be unrelated and purely coincidental, but there have been 5 leap seconds added to UTC since J2000.

haileyajohnson:
@JohnLCaron  this was introduced in [this commit](https://github.com/Unidata/netcdf-java/commit/4825d27fc60b04b5b5f3d10b61bddbfbf7fccba8), do you happen to remember what was going on then?

PR #1005, opened 2022-04-27
[5.x] remove 5 sec slop from getOffset

haileyajohnson:
- remove extra five seconds from `CalendarPeriod.getOffset`


haileyajohnson:
Potentially address #980, unless we know why the 5 seconds are being added

XSalard:
You might want to propagate this fix to 6.x and 7.x too. 
It affects cdm-core/src/main/java/ucar/nc2/time/CalendarPeriod.java in the exact same way.

tdrwenski (approved):

PR #1007, opened 2022-04-27
[5.x] use 'release' compiler flag

haileyajohnson:
- compile using the `release` flag instead of source/target if Java version > 8


haileyajohnson:
This simplicity of this solution is straight up infuriating.
Does not incorporate toolchains for build/test with different versions - still on the TODO list, but not imminently important now.

tdrwenski (approved):
That's smart! That solution didn't occur to me😆

PR #1008, opened 2022-04-28
Make sure that DAP4 xml strings do not contain NUL characters.

DennisHeimbigner:
The current remotetest/d4ts servlet is failing because it is including the trailing NUL characters in the DMR XML metadata. This is illegal XML.  Fix, of course, is to suppress NUL characters.

This change has been tested on remotetest, but it is only indirectly tested in the existing DAP4 tests.

This is the first in a probably long list of DAP4 fixes.


haileyajohnson (commented):
Needs `./gradlew spotlessJavaApply`

dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  //////////////////////////////////////////////////
+  // Junit test method
+
+  @Test
+  public void testDAP4Client() throws Exception {
+    boolean pass = true;
+    for (ClientTest testcase : chosentests) {
+      if (!doOneTest(testcase))
```
haileyajohnson, line 291: I think at some point we'll want to refactor this to use a Parameterized JUnit test, for better granularity on the test results (and to match the parts of the project that have been updated more recently)

Issue #997, opened 2022-04-14
[5.5.2]: Issues reading compressed netCDF files with toolsUI 5.5.2 via nj22Config.xml 


mcucchi9:
### Description

I'm trying to read a JPEG-LS compressed netCDF file with ToolsUI 5.5.2 using a decompression library written in C, by allowing the use of the netCDF-C library for reading purposes through the `$HOME\.unidata\nj22Config.xml` configuration file (https://github.com/Unidata/thredds/issues/1063). 
While I am able to do it (and also perform a decompression through the "Writer" tab functionalities) with v5.1.0, I can't with the latest version: as soon as I open the compressed file I get the following warning message on my terminal:

```
[16:57:42.645 WARN  u.nc2.internal.iosp.hdf5.H5headerNew]   Variable effective_radiance has unknown Filter(s) =    MessageFilter filters=
    Filter id= 32018 flags = 1 nValues=13 name= HDF5 JPEG-LS filter data = 2 1 279 11136 16 1 0 0 0 0 0 0 0 
```

A similar issue applies to Panoply 5.0.2 (based upon netcdfAll-5.5.2) with respect to Panoply 4.10.12 (based upon netcdfAll-5.1.0): while I am able to open and plot variables of JPEG-LS compressed files with v4.10.12 exploiting the `$HOME\.unidata\nj22Config.xml` file, I can't with v5.0.2, getting the following error message when trying to create a plot:

```
There was an exception preparing the data: Unknown filter type=32018
```

### Environmental information

**OS**: Ubuntu 20.04.4 LTS
**Java version**: openjdk version "13.0.7" 2021-04-20

#### Other details

- I'm working in a ``conda`` environment where the decompression filter is correctly configured to be used by the netCDF-C library (tested with netCDF-C tools such as ``nccopy``)
- the ``$HOME/.unidata/nj22Config.xml`` file is configured as follows
   ```
   <nj22Config>
     <Netcdf4Clibrary>
        <libraryPath>$NETCDFC_LIB_PATH</libraryPath>
        <libraryName>netcdf</libraryName>
        <useForReading>true</useForReading>
     </Netcdf4Clibrary>
   </nj22Config>
   ```
   where `$NETCDFC_LIB_PATH` is the explicit full path to the ``conda`` environment ``lib`` directory containing the netCDF-c library `libnetcdf.so`
- I tried to launch the mentioned Java applications both with
   ```
   java -jar <full_path_to_the_java_application_jar>
   ```
[...]

PR #1009, opened 2022-04-28
Fix issue #997

cofinoa:
This PR fixes issue #997.


haileyajohnson (approved):
Thanks @cofinoa !

lesserwhirls:
We need to keep the calls to the old API (`NetcdfFile.registerIOProvider`), otherwise the old API won't be able to use the netCDF-C library for reading.

PR #1010, opened 2022-04-28
[5.x] Some cleanups

tdrwenski:
- refactor string parsing logic for CollectionSpecParser
- cleanup a few tests

### DIFF
--- a/cdm/core/src/main/java/thredds/inventory/CollectionSpecParser.java
+++ b/cdm/core/src/main/java/thredds/inventory/CollectionSpecParser.java
@@ -62,62 +62,16 @@ public class CollectionSpecParser {
    */
   public CollectionSpecParser(String collectionSpec, Formatter errlog) {
     this.spec = collectionSpec.trim();
-    int posFilter;
 
-    int posGlob = collectionSpec.indexOf("/**/");
-    if (posGlob > 0) {
-      rootDir = collectionSpec.substring(0, posGlob);
-      posFilter = posGlob + 3;
-      subdirs = true;
-
-    } else {
-      subdirs = false;
-      posFilter = collectionSpec.lastIndexOf('/');
-      if (posFilter > 0)
-        rootDir = collectionSpec.substring(0, posFilter);
-      else
-        rootDir = System.getProperty("user.dir"); // working directory
-    }
+    rootDir = getRootDir(collectionSpec);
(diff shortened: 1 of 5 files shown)


cdm/core/src/main/java/thredds/inventory/CollectionSpecParser.java
```diff
@@ -132,6 +86,67 @@ public CollectionSpecParser(String rootDir, String regExp, Formatter errlog) {
...
+      return Pattern.compile(filterAndDateMark);
+    } else if (numberOfHashes == 1) {
+      return Pattern.compile(filterAndDateMark.substring(0, filterAndDateMark.indexOf('#')) + "*");
+    } else {
+      StringBuilder sb = new StringBuilder(StringUtil2.remove(filterAndDateMark, '#')); // remove hashes, replace with .
+
+      for (int i = filterAndDateMark.indexOf('#'); i < filterAndDateMark.lastIndexOf('#') - 1; i++) {
+        sb.setCharAt(i, '.');
+      }
+
```
haileyajohnson: can this just be `filterAndDateMark.replace("#", ".")`?

cdm/core/src/test/java/thredds/inventory/TestCollectionSpecParser.java
```diff
@@ -1,45 +1,105 @@
...
 import java.util.Formatter;
+import java.util.List;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-/** Sanity check on CollectionSpecParser */
+@RunWith(Parameterized.class)
```
haileyajohnson, line 16: 😃

haileyajohnson (approved):

cdm/core/src/main/java/thredds/inventory/CollectionSpecParser.java
tdrwenski: I think that comment on line 124 is a bit misleading. We want to replace the whole date part with  "."-- So if you have `directory/foo#yyyyMMdd_HHmm#.*grib1` it assumes the part between the first # and the last # is the date and replaces that with periods in the filter to make `regEx=foo..............*grib1`

didn't see a one liner but I tried to make it more readable in 9fa6453c7f

PR #1012, opened 2022-05-02
[5.x] S3 collection spec parsing

tdrwenski:
- S3 version of the collection spec parser with an abstract base class
- Tests for the S3 collection spec parsing
- Fix MFileCollectionManager for S3 collection specs
- Add tests for MFileCollectionManager, local and s3, to verify files given by spec are being correctly found

### DIFF
--- a/cdm/core/src/test/java/thredds/inventory/TestCollectionSpecParser.java
+++ b/cdm/core/src/test/java/thredds/inventory/TestCollectionSpecParser.java
@@ -58,10 +58,10 @@ public class TestCollectionSpecParser {
         {"/data/ldm/pub/decoded/netcdf/surface/metar/**/", "/data/ldm/pub/decoded/netcdf/surface/metar", true, null,
             null},
 
-        {"/data/ldm/pub/decoded/netcdf/surface/metar/**/*", "/data/ldm/pub/decoded/netcdf/surface/metar", true, null,
+        {"/data/ldm/pub/decoded/netcdf/surface/metar/**/.*", "/data/ldm/pub/decoded/netcdf/surface/metar", true, ".*",
             null},
 
-        {"/data/ldm/pub/decoded/netcdf/surface/metar/*", "/data/ldm/pub/decoded/netcdf/surface/metar", false, null,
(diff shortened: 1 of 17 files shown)


cdm/core/src/test/java/thredds/inventory/TestCollectionSpecParser.java
```diff
@@ -58,10 +58,10 @@ public static List<Object[]> getTestParameters() throws URISyntaxException {
         {"/data/ldm/pub/decoded/netcdf/surface/metar/**/", "/data/ldm/pub/decoded/netcdf/surface/metar", true, null,
             null},
 
-        {"/data/ldm/pub/decoded/netcdf/surface/metar/**/*", "/data/ldm/pub/decoded/netcdf/surface/metar", true, null,
+        {"/data/ldm/pub/decoded/netcdf/surface/metar/**/.*", "/data/ldm/pub/decoded/netcdf/surface/metar", true, ".*",
```
tdrwenski: I very slightly changed the behavior of the original `CollectionSpecParser`. Previously, if the regEx was only one character (here '\*' but could be any character) then it resulted in `filter = null`. Now, a regex of '*' will be compiled into a `Pattern` and will fail.

haileyajohnson: hmm, I'm a little worried that will break some users, I think the single '*' pattern is pretty commonly used...

tdrwenski: It just sort of seemed like a mistake that a one character filter was ignored. You can achieve the same results with no filter or with ".*".

But regardless that could be annoying for users already using a "*" so I can change it back.

haileyajohnson: yeah, I think '**/*' is pretty standard for "anything at any nesting level", even if it is unnecessary

tdrwenski: back to original behavior in 35143be7db

haileyajohnson (approved):

PR #1014, opened 2022-05-02
[5.x] Add methods to mfile

tdrwenski:
Add `exists` and `getInputStream` methods to MFile. This is needed so that an S3 `MFile`s can be used for grib indexes. The `exists` method will be useful in checking whether a grib index already exists or needs to be created. The `getInputStream` method is needed for reading a grib index.

### DIFF
--- a/cdm/core/src/test/java/thredds/filesystem/TestMFileOS.java
+++ b/cdm/core/src/test/java/thredds/filesystem/TestMFileOS.java
@@ -5,66 +5,94 @@ import static com.google.common.truth.Truth.assertThat;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.util.Arrays;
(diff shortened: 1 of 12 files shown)


cdm/core/src/test/java/thredds/filesystem/TestMFileOS.java
```diff
@@ -11,60 +11,81 @@
 import java.util.Random;
 import org.junit.ClassRule;
 import org.junit.Test;
+import org.junit.experimental.runners.Enclosed;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
+import thredds.inventory.MFile;
 
-@RunWith(Parameterized.class)
```
tdrwenski, line 18: If you "hide whitespace changes" this diff is more readable-- I wrapped the parameterized tests in a inner class so I could run unparameterized tests as well

haileyajohnson (approved):

PR #1017, opened 2022-05-11
fix bug in ncss with projection coords

haileyajohnson:
- fix rounding error causing exception in ncss

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoordAxisHelper.java
+++ b/cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoordAxisHelper.java
@@ -86,13 +86,14 @@ class CoordAxisHelper {
 
     double distance = coordValue - axis.getCoordEdge1(0);
     double exactNumSteps = distance / axis.getResolution();
-    // int index = (int) Math.round(exactNumSteps); // ties round to +Inf
-    int index = (int) exactNumSteps; // truncate down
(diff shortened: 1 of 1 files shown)


cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoordAxisHelper.java
```diff
@@ -87,7 +87,7 @@ private int findCoordElementRegular(double coordValue, boolean bounded) {
     double distance = coordValue - axis.getCoordEdge1(0);
     double exactNumSteps = distance / axis.getResolution();
     // int index = (int) Math.round(exactNumSteps); // ties round to +Inf
-    int index = (int) exactNumSteps; // truncate down
+    int index = exactNumSteps > 0 ? (int) exactNumSteps : (int) exactNumSteps - 1; // truncate down
```
tdrwenski: isn't this the same as using `floor`? I thought casting to int rounds towards zero and floor always truncates down which I think is what you want here.

tdrwenski (approved):

cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoordAxisHelper.java
haileyajohnson: you right, brb

PR #1021, opened 2022-05-31
[5.x] Fix projection coordinates bugs

tdrwenski:
From support ticket GTP-562460.

- Fix HorizCoordSys handling of units. The projection code always assumes km, so need to convert coordinate values to/from km. This is a hacky fix-- I don't see a better place to handle this nicely, but if you have ideas I will have another look.
- Fix GridCoordSys calculation of latitude/longitude bounding box from projection coordinates bounding box. Previously, it only used the corners of the projection coordinates bounding box, leading to cut off data in WMS.
- Add tests


cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
```diff
@@ -646,12 +654,37 @@ private List<LatLonPoint> calcLatLonBoundaryPointsFromProjection(int maxPointsIn
...
+      return coordinate;
+    }
+  }
+
+  private static double convertFromKm(double coordinate, String unit, String axisName) {
+    if (unit.equals("km") || unit.equals("kilometers")) {
+      return coordinate;
+    } else if (unit.equals("m") || unit.equals("meters")) {
+      return 1000 * coordinate;
+    } else {
```
haileyajohnson, line 683: isn't this backwards, i.e. unit=km should return 1000*coordinate?

cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
```diff
@@ -646,12 +654,37 @@ private List<LatLonPoint> calcLatLonBoundaryPointsFromProjection(int maxPointsIn
...
   }
 
+  // TODO is there a better place to handle units?
+  private static double convertToKm(double coordinate, String unit, String axisName) {
+    if (unit.equals("km") || unit.equals("kilometers")) {
+      return coordinate;
+    } else if (unit.equals("m") || unit.equals("meters")) {
+      return 0.001 * coordinate;
+    } else {
+      // TODO not sure what other units that may not need scaling could be here (e.g. radians)
```
haileyajohnson: maybe for now just log that it's an "unrecognized unit"?

haileyajohnson (changes requested):

cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
haileyajohnson: or are the value/units mismatched in this case?

tdrwenski: Think it is correct but maybe the variable/ function names could be improved-- the coordinate in this case is a computed one in km, and is returned from using the projection to convert from la/lon, i.e. from `proj.latLonToProjBB(llbb)` and the unit is the unit that the data is actually in, e.g. meters.

Maybe `convertFromKm(double coordinateInKm, String desiredUnit, String axisName)` ?

haileyajohnson (approved):

cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
tdrwenski: added a log message in 571106c23f-- only made it "info" level because if the units are radians for example then there is no problem.

cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
```diff
@@ -675,14 +675,14 @@ private static double convertToKm(double coordinate, String unit, String axisNam
...
-  private static double convertFromKm(double coordinate, String unit, String axisName) {
-    if (unit.equals("km") || unit.equals("kilometers")) {
-      return coordinate;
-    } else if (unit.equals("m") || unit.equals("meters")) {
-      return 1000 * coordinate;
+  private static double convertFromKm(double coordinateInKm, String desiredUnit, String axisName) {
+    if (desiredUnit.equals("km") || desiredUnit.equals("kilometers")) {
+      return coordinateInKm;
+    } else if (desiredUnit.equals("m") || desiredUnit.equals("meters")) {
+      return 1000 * coordinateInKm;
```
haileyajohnson, line 682: yeah this is a lot clearer :)

haileyajohnson (approved):

Issue #1018, opened 2022-05-20
Issue in CoordTransformBuilder related to UTM class registration


ghansham:
The name with which UTM.class is registered here:

https://github.com/Unidata/netcdf-java/blob/a49395075f29f87defacc84950bb7c9a3a5653f0/cdm/core/src/main/java/ucar/nc2/dataset/CoordTransBuilder.java#L57

is different from the value which is returned by this method:
https://github.com/Unidata/netcdf-java/blob/a49395075f29f87defacc84950bb7c9a3a5653f0/cdm/core/src/main/java/ucar/nc2/dataset/transform/UTM.java#L16

Issue comes here is that if we create a file with grid_mapping_name set to 'universal_transverse_merctator', netcdf-java does not recognizes its coordinate system.
This bug is available in netCDF-4.6.X and netCDF-5.X.

PR #1029, opened 2022-06-10
[5.x] register UTM by UtmProjection grid mapping name

haileyajohnson:
## Description of Changes

- add "universal_transverse_mercator" to transform map
- see #1018 

## PR Checklist

       until ready for review

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/dataset/CoordTransBuilder.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/CoordTransBuilder.java
@@ -55,6 +55,7 @@ public class CoordTransBuilder {
     registerTransform(CF.STEREOGRAPHIC, Stereographic.class);
     registerTransform(CF.TRANSVERSE_MERCATOR, TransverseMercator.class);
     registerTransform("UTM", UTM.class);
+    registerTransform("universal_transverse_mercator", UTM.class);
     registerTransform(CF.VERTICAL_PERSPECTIVE, VerticalPerspective.class);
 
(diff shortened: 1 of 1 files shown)


cdm/core/src/main/java/ucar/nc2/dataset/CoordTransBuilder.java
```diff
@@ -55,6 +55,7 @@ public class CoordTransBuilder {
     registerTransform(CF.STEREOGRAPHIC, Stereographic.class);
     registerTransform(CF.TRANSVERSE_MERCATOR, TransverseMercator.class);
     registerTransform("UTM", UTM.class);
+    registerTransform("universal_transverse_mercator", UTM.class);
```
tdrwenski, line 58: should this string be added to the constants in `CF`?

tdrwenski (approved):

cdm/core/src/main/java/ucar/nc2/dataset/CoordTransBuilder.java
haileyajohnson: I think that list comes from [here](https://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#appendix-grid-mappings) and UTM isn't in there.

ghansham: > should this string be added to the constants in `CF`?

It's not there in CF conventions but there should be consistency between default names for transform classes are returned and the one being used to register the classes within Unidata CDM.

Issue #1023, opened 2022-06-03
5.5.2: CoordinateAxis1DTime::getCoordBoundsDate returns nonsensical bounding dates when axis size equals 1


tombrauer:
### Versions impacted by the bug

v5.x

### What went wrong?

`CoordinateAxis1DTime::getCoordBoundsDate` returns nonsensical bounding dates when `CoordinateAxis1DTime::getSize` is equal to 1. It seems like the logic in `CoordinateAxis1DTime::getCoordBoundsDate` is predicated on the axis size being greater than 1. Should my code check axis size == 1 before calling `CoordinateAxis1DTime::getCoordBoundsDate` or could the logic in `CoordinateAxis1DTime::getCoordBoundsDate` be more robust? 

Remove the txt extension from example files below.
[NWRFC_2017-12-31T2300_to_2018-01-01T0000.nc4.txt](https://github.com/Unidata/netcdf-java/files/8835329/NWRFC_2017-12-31T2300_to_2018-01-01T0000.nc4.txt)
[01.nc.txt](https://github.com/Unidata/netcdf-java/files/8835334/01.nc.txt)
 

### If you have an example file that you can share, please attach it to this issue.
If so, may we include it in our test datasets to help ensure the bug does not return once fixed?
Note: the test datasets are publicly accessible without restriction.

Yes

### Code of Conduct

- [X] I agree to follow the UCAR/Unidata Code of Conduct

haileyajohnson:
What behavior would you expect/want if the size of the axis=1? There aren't technically any bounds on the coordinate if there is only one coordinate.
These are the options as I see it:

- return null
- return an array of size two with the same, single coordinate repeated twice
- throw an exception

We're open to other suggestions of how to handle it as well.

tombrauer:
I ran the same test with a grib file and curiously that behaves differently than netCDF (maybe due to the underlying IOSP implementation?). The grib file behaves like the second approach you describe by returning an array of size two with the same, single coordinate repeated twice. See sample file below. 

[MRMS_GaugeCorr_QPE_01H_00.00_20170102-120000.grib2.txt](https://github.com/Unidata/netcdf-java/files/8874223/MRMS_GaugeCorr_QPE_01H_00.00_20170102-120000.grib2.txt)

The second approach is what works best for my code, but if you decide to handle in another way, I can add appropriate handling logic.

PR #1030, opened 2022-06-10
handle get coord bounds for an axis with size=1

haileyajohnson:
## Description of Changes

see #1023 

## PR Checklist

       until ready for review

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/dataset/CoordinateAxis1D.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/CoordinateAxis1D.java
@@ -976,7 +976,7 @@ public class CoordinateAxis1D extends CoordinateAxis {
       return;
     }
     if (size == 1) {
-      // if the axis size is zero, edges are the same as the coord
+      // if the axis size is one, edges are the same as the coord
       edge[0] = coords[0];
       edge[1] = coords[0];
       return;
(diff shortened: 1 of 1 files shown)


cdm/core/src/main/java/ucar/nc2/dataset/CoordinateAxis1D.java
```diff
@@ -972,10 +972,18 @@ private boolean makeBoundsFromAux() {
   private void makeEdges() {
     int size = (int) getSize();
     edge = new double[size + 1];
-    if (size < 1)
+    if (size < 1) {
       return;
-    for (int i = 1; i < size; i++)
+    }
+    if (size == 1) {
+      // if the axis size is zero, edges are the same as the coord
```
tdrwenski: do you mean if the axis size is one?

cdm/core/src/main/java/ucar/nc2/dataset/CoordinateAxis1D.java
```diff
@@ -972,10 +972,18 @@ private boolean makeBoundsFromAux() {
...
+      edge[0] = coords[0];
+      edge[1] = coords[0];
+      return;
+    }
+    for (int i = 1; i < size; i++) {
       edge[i] = (coords[i - 1] + coords[i]) / 2;
+    }
     edge[0] = coords[0] - (edge[1] - coords[0]);
     edge[size] = coords[size - 1] + (coords[size - 1] - edge[size - 1]);
     isContiguous = true;
```
tdrwenski, line 989: should this be set to true in the case `size == 1`? The default is also true so not sure if it is needed

tdrwenski (approved):

cdm/core/src/main/java/ucar/nc2/dataset/CoordinateAxis1D.java
haileyajohnson: I don't think it matters when there's just one point. Contiguous is meaningless then, and (I think) we're handling it further up the pipeline well enough that it works.

PR #1051, opened 2022-08-03
[5.x] Fix horizontal subset for all longitudes

tdrwenski:
## Description of Changes

This fixes the problem with the longitudes shown in the ncss horizontal extent being the same for east and west when the all longitudes are included. Not sure if this is the best way to fix this, and the mouse over still shows non-normalized values (for instance 179 to 539) I think because of the way the connected longitude/latitudes are calculated.

## PR Checklist

       until ready for review

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
+++ b/cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
@@ -493,7 +493,7 @@ public class HorizCoordSys {
       maxLon = Math.max(maxLon, boundaryPoint.getLongitude());
     }
 
-    return new LatLonRect(LatLonPoint.create(minLat, minLon), LatLonPoint.create(maxLat, maxLon));
+    return new LatLonRect(LatLonPoint.create(minLat, minLon), maxLat - minLat, maxLon - minLon);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
```diff
@@ -493,7 +493,7 @@ public LatLonRect calcLatLonBoundingBox() {
       maxLon = Math.max(maxLon, boundaryPoint.getLongitude());
     }
 
-    return new LatLonRect(LatLonPoint.create(minLat, minLon), LatLonPoint.create(maxLat, maxLon));
+    return new LatLonRect(LatLonPoint.create(minLat, minLon), maxLat - minLat, maxLon - minLon);
```
tdrwenski, line 496: The `LatLonPoint.create` normalizes the longitude to between -180 and +180 (not sure if this is good but it seemed difficult to change). So previously, the constructed `LatLonRect` would have same longitude for the left and right of the rectangle (for instance -181 and 179 would both become 179). Using the other constructor (with a deltaLongitude) allows the `LatLonRect` to know that the `width` of the rectangle is 360 instead of 0.

haileyajohnson (approved):

Issue #1052, opened 2022-08-04
Can't bypass scaling in Grayscale Geotiff Writing


WeatherGod:
https://github.com/Unidata/netcdf-java/blob/b3a3726ca760534ffae7154377ccc80443609eb1/cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java#L582

My data that is being served by the WCS service in TDS is already unsigned bytes for categorical data, but the geotiff I get from WCS gets scaled and loses the original categorical values. It would be nice to have a way to either flag that the data should not get scaled/offset prior to saving, or detect that the input/output grid dtypes are the same and leave things alone?

Even better would be to have an option for paletted geotiffs, which my source dataset has specified list of colors that works for the WMS service.

WeatherGod:
Closed by #1066

PR #1066, opened 2022-08-22
[5.x]: Adding the ability to save geotiffs with a palette

WeatherGod:
## Description of Changes

Add support to save out a geotiff with supplied colortable

## PR Checklist

       until ready for review


WeatherGod:
This is my first significant chunk of Java code, so don't assume I did any of this correctly.

The reason for this submission and its design is that I plan to also make a submission to TDS's WCS service to add a new Geotiff_Palette format that would then respect a dataset's "flag_values" and "flag_colors" attributes, much like how ncWMS does right now. We could also respect supplied styling options this way, too.

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
```diff
@@ -213,8 +215,16 @@ private void writeMetadata(boolean greyScale, double xStart, double yStart, doub
...
       geotiff.addTag(new IFDEntry(Tag.YResolution, FieldType.RATIONAL).setValue(1, 1));
       geotiff.addTag(new IFDEntry(Tag.ResolutionUnit, FieldType.SHORT).setValue(1));
-
+      if (colorTable != null && colorTable.length > 0) {
+        geotiff.addTag(new IFDEntry(Tag.PhotometricInterpretation, FieldType.SHORT).setValue(3));
+        geotiff.addTag(new IFDEntry(Tag.ColorMap, FieldType.SHORT, colorTable.length).setValue(colorTable));
+      } else {
+        // black is zero (value used in GeotiffWriter)
+        geotiff.addTag(new IFDEntry(Tag.PhotometricInterpretation, FieldType.SHORT).setValue(1));
+      }
```
WeatherGod: Probably need to do a bit of thinking about this method's call signature, particularly the `greyscale` argument. Need to consider how orthogonal these features are.

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
```diff
@@ -261,6 +271,91 @@ private void writeMetadata(boolean greyScale, double xStart, double yStart, doub
...
+   */
+  public void setColorTable(Map<Integer, Color> colorMap, Color defaultRGB) {
+    if (colorMap == null) {
+        colorTable = null;
+        return;
+    }
+
+    // FIXME: This isn't quite right because this assumes that the data being written is
+    // unsigned bytes, but the tiff spec says that it should be sized to the width of
+    // the data type, but I would need to know the data type, which this writer doesn't know.
```
WeatherGod: Might also want to consider optional support for the alpha channel? I haven't tried making a paletted geotiff with a fourth channel yet.

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
```diff
@@ -261,6 +271,91 @@ private void writeMetadata(boolean greyScale, double xStart, double yStart, doub
...
+      for (int i = 0; i < 256; i++) {
+        // Scale it up to [0, 65535], which is needed by the ColorMap tag.
+        if (color == 0) {
+          colorTable[i] = colorMap.getOrDefault(i, defaultRGB).getRed() * 256;
+        } else if (color == 1) {
+          colorTable[256 + i] = colorMap.getOrDefault(i, defaultRGB).getGreen() * 256;
+        } else {
+          colorTable[512 + i] = colorMap.getOrDefault(i, defaultRGB).getBlue() * 256;
+        }
+      }
```
WeatherGod: Maybe I should unroll the color loop?

WeatherGod:
And, this is partly in response to https://github.com/Unidata/netcdf-java/issues/1052

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
WeatherGod: Yeah, this isn't working out too great conceptually (and it is making patching difficult in tds). Since this is a private method, can I take that to mean I can change up the call signature a bit? Basically, I'm thinking of adding a new argument `dtype` that can specify what `DataType` the output should be (or should I use `FieldType`?). If it is `null`, then it means "auto" and it'll match the dtype of the given data.

For backwards compatibility, I could ignore `dtype` if greyscale is true.

Thoughts? It'll require a new writeData() for integer type (there is currently one for byte and float). Probably wouldn't cover everything geotiff can handle, but should be pretty decent, right?

haileyajohnson: yes,  you can change the signature for this method since it is private.

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
haileyajohnson: it seems like you don't need the outer loop here?

WeatherGod: You do. The TIFF specification for the colortable is basically like a fortran-ordered array, so all of the reds first, then greens, then blues.

haileyajohnson: But you're getting all three colors at once in the inner loop?

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
WeatherGod: Great! I kind of want to split this function up a bit so that the "greyscale" stuff is out of here and can be handled separately. It would make this function and my proposed `dtype` approach much more sane and reasonable. We can then relegate the `greyscale` handling as a special-case that doesn't care about any of the dtype information.

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
haileyajohnson: Oh nvm, you are getting them one at a time. But you don't have to be? You could remove the outer loop and the inner if statements entirely

WeatherGod: True! I was so focused on filling up the array sequentially. No clue if there are performance issues either way.

WeatherGod:
I am just about done with a set of revisions that I think makes things more sane, but I needed to write a `coerceInt()` and a `coerceByte()` private methods to coerce an arbitrary `Array` into `ArrayInt` or `ArrayByte` copies, respectively. I feel like that is some sort of NiH syndrome. Is there already a "correct" way to do what in `numpy` would be like an `a.astype(int)`?

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
WeatherGod: Thinking some more on this FIXME of mine. The only place where one knows what the dtype is going to be is in `writeGrid()` (and consequently, `writeMetadata()`). The GeotiffWriter class acts more like a state-less methods class rather than some sort of stateful "builder" class. Maybe my `setColorTable()` approach is an anti-pattern here and I really should be making a `ColorTable` class that can get passed into `writeGrid()` and `writeMetadata()`? Then, when the time comes to construct the colortable array, I'll know the width of the output dtype and can construct it appropriately?

cdm/misc/src/test/java/ucar/nc2/geotiff/TestGeoTiffPalette.java
```diff
@@ -0,0 +1,162 @@
...
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.util.HashMap;
+import java.awt.Color;
+
+/**
+ * Testing the color palette features of GeotiffWriter
+ *
+ * @author WeatherGod
+ * @since 8/12/20224
```
WeatherGod: whoops

cdm/misc/src/test/java/ucar/nc2/geotiff/TestGeoTiffPalette.java
```diff
@@ -0,0 +1,162 @@
...
+    GeotiffWriter writer = new GeotiffWriter("dummy.tif");
+    writer.setColorTable(colorMap, Color.black);
+    int[] resultTable = writer.getColorTable();
+
+    // Explicitly spelling out the RGB channel values [0, 255]
+    // because there were some off-by-one errors in initial development.
+    int[] expectedRGBs = new int[] {0, 170, 255,  // #00AAff
+                                    21, 20, 18,   // #151412
+                                    222, 1, 171,  // #DE01aB
+                                    16, 10, 187}; // #100ABB
```
WeatherGod: So, spotless wants to re-indent this in a not-ideal manner. Are there any suggested formatting options I can utilize, or am I stuck with these last three lines being out of alignment with the first?

cdm/misc/src/main/java/ucar/nc2/geotiff/GeoTiff.java
```diff
@@ -191,6 +192,30 @@ int writeData(byte[] data, int imageNumber) throws IOException {
     return nextOverflowData;
   }
 
+  int writeData(int[] data, int imageNumber) throws IOException {
```
WeatherGod, line 220: Untested at the moment

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
```diff
@@ -144,20 +210,29 @@ void writeGrid(GridDatatype grid, Array data, boolean greyScale, double xStart,
...
+      if (dtype == DataType.BYTE || dtype == DataType.UBYTE) {
+        ArrayByte result = coerceByte(data);
+        nextStart = geotiff.writeData((byte[]) result.getStorage(), imageNumber);
+      } else if (dtype.isIntegral()) {
+        ArrayInt result = coerceInt(data);
+        nextStart = geotiff.writeData((int[]) result.getStorage(), imageNumber);
+      } else {
+        ArrayFloat result = replaceMissingValues(grid, data, dataMinMax);
+        nextStart = geotiff.writeData((float[]) result.getStorage(), imageNumber);
+      }
```
WeatherGod: I should see what I can do about code deduplication, as this is nearly identical to code in another `writeGrid()` method.

WeatherGod: Should also see about adding some unit tests to make sure that the data is getting saved correctly in different formats and coerced correctly.

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
```diff
@@ -261,6 +381,117 @@ private void writeMetadata(boolean greyScale, double xStart, double yStart, doub
...
+   * Have the geotiff include a colormap in the form of a mapping of the pixel value to the rgb triplet.
+   * Provide a default RGB triplet for any values not specified.
+   *
+   * Pass null to unset the colorTable.
+   *
+   * For these RGB triplets, 0 is minimum intensity, 255 is maximum intensity. Values outside that range will
+   * be floored/ceilinged to the [0, 255] range. The color table is also assumed to be for pixel values
+   * between 0 and 255.
+   * In order for the color table to be properly included in the geotiff, the "greyScale" mode must be false,
+   * and the output data type must be byte or integer.
```
WeatherGod: I should actually just restrict this to ubyte or add tests to verify that byte and/or integers work (which I doubt).

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
```diff
@@ -213,17 +300,54 @@ private void writeMetadata(boolean greyScale, double xStart, double yStart, doub
...
+          ftype = FieldType.SHORT;
+        } else if (sdtype == DataType.INT) {
+          // A geotiff LONG is really a 4-byte regular integer
+          ftype = FieldType.LONG;
+        } else {
+          throw new IllegalArgumentException("Unsupported dtype: " + dtype + " sdtype: " + sdtype);
+        }
+        geotiff.addTag(new IFDEntry(Tag.SMinSampleValue, ftype).setValue(min));
+        geotiff.addTag(new IFDEntry(Tag.SMaxSampleValue, ftype).setValue(max));
+        geotiff.addTag(new IFDEntry(Tag.GDALNoData, ftype).setValue(min - 1));
```
WeatherGod: Just discovered an odd bug here. When the original data's range goes from 0 to 255 for ubyte type, this attribute gets set to 0, which messes up the color table since I wanted 255 to be NoData. At the moment, my guess is that a normalization elsewhere is making it that 1 is always the minimum and therefore 0 always becomes NoData. So, I might need to double-check whether or not normalization is even desired in the first place and allow other mechanisms for specifying a NoData attribute.

tdrwenski:
Hi @WeatherGod, are you still working on this PR?

WeatherGod:
Yes in the sense that it is a feature for a client of mine, so I've been
maintaining it for them. There is a bug of sorts, but I think it is
actually pre-existing. I would appreciate feedback on what I have so far
and if there is anything else you all would like added to it.

tdrwenski:
Okay, thanks! Will try to have a look at your comments about the possible bug. To see the style issues you can run `./gradlew spotlessCheck` or to automatically fix them you can run `./gradlew spotlessApply`

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
```diff
@@ -213,17 +300,54 @@ private void writeMetadata(boolean greyScale, double xStart, double yStart, doub
...
+        geotiff.addTag(new IFDEntry(Tag.SampleFormat, FieldType.SHORT).setValue(3)); // IEEE Floating Point Type
+      } else {
+        throw new IllegalArgumentException("Unsupported data type for geotiff: " + dtype);
+      }
       geotiff.addTag(new IFDEntry(Tag.SamplesPerPixel, FieldType.SHORT).setValue(1));
-      float min = (float) (dataMinMax.min);
-      float max = (float) (dataMinMax.max);
-      geotiff.addTag(new IFDEntry(Tag.SMinSampleValue, FieldType.FLOAT).setValue(min));
-      geotiff.addTag(new IFDEntry(Tag.SMaxSampleValue, FieldType.FLOAT).setValue(max));
-      geotiff.addTag(new IFDEntry(Tag.GDALNoData, FieldType.FLOAT).setValue(min - 1.f));
```
WeatherGod: Just realized that this might be technically wrong. According to https://www.awaresystems.be/imaging/tiff/tifftags/gdal_nodata.html, this field is supposed to be ASCII encoded?

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
WeatherGod: Alright, I think I understand what the bug is. Over at https://github.com/Unidata/netcdf-java/pull/1066/files#diff-b37a37fa215c4b41463e379358f9e9748244261adee3067e72fea3e231344aeaR851, when not writing out greyscale images, I made it so that writing out integer, short, or byte data is done without scaling and missing data handling. So, I shouldn't be setting this field here with "min - 1" since scaling of the original data isn't happening. Instead, I probably should be checking the `grid` object to find out what its missing data setting is if any and use that, maybe?

WeatherGod: Hmmm, seems like the underlaying VariableDS object returns a list of missing values rather than a single definitive one. Is there an accepted approach to determine a dataset's "missing value"? In xarray & python-netcdf4, it seems like it is usually set as the dataset's "_FillValue" attribute, but I have also seen it set as the "missing" attribute.

haileyajohnson: A variable can have multiple missing values, in addition to a fill value, If your data only has one missing values, `getMissing` will return an array of just that value.

WeatherGod: Right, I'm just trying to figure out what should be considered "the right/best way" since I can only encode a single NoData value. Like, are the values in the list ordered by "priority" or something? Maybe it would be best to not guess at the user's intent and not encode this at all?

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
WeatherGod: Should I fix this as part of this PR or submit a separate PR? I'm just now realizing that this mistake has been causing all sorts of low-level "weird" issues that I could never quite figure out.

WeatherGod: Just came to the realization that there are no pre-existing tests for the geotiff writer with greyscale set to false. I think I'm going to take care of that first along with this nodata fix in a separate PR.

WeatherGod: Opened https://github.com/Unidata/netcdf-java/pull/1335 to address the NoData issue and to expand the tests a bit. I'll rebase this PR on top of that.

cdm/misc/src/main/java/ucar/nc2/geotiff/GeoTiff.java
WeatherGod: Any suggestions for a test datafile that has integer variables? I found `ft/problem/88129.nc`, but I can't seem to write the data back out properly (I need to investigate the test a little bit further on that).

WeatherGod: I completely forgot that my test file has an integer dtype in it. And utilizing this has revealed some other possible bugs that I'm now going to hunt down (possibly a bug with the test rather than the code).

WeatherGod: Figured out the test failure and there is a bug in the geotiff writer when doing lambert conformal and the y-axis goes from positive to negative. I'll open up a new PR for that.

WeatherGod: Opened https://github.com/Unidata/netcdf-java/pull/1342 to address this. I will need to rebase this PR to resolve the conflict since the test file I introduce here will be making an early appearance.

WeatherGod:
Crazy thought,, but what if we allowed color tables to work with `greyscale==true`? Right now, as coded, greyscale mode scales the given data to the range of 1-255 and encodes missing data as 0, effectively unsigned bytes, while the "colortable" mode doesn't do any scaling at all, and only works if the data type is unsigned bytes. But, it is getting tedious to document and code checks that the colortable can only be used when greyscale is false. Why not allow color tables to be used for greyscale mode?

It might be easier to explain the various features available as "greyscale normalizes to UBYTE", "color table works only for UBYTEs", and "Non-greyscale FLOATs encodes missing data as data minimum minus one".

Thoughts?

WeatherGod:
Now that I've actually implemented that idea, I really like it. I'm running with it. It also addresses several other questions I had in my mind about possible use-cases and whether or not the setColorTable() was an anti-pattern.

cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
```diff
@@ -213,17 +331,55 @@ private void writeMetadata(boolean greyScale, double xStart, double yStart, doub
...
       geotiff.addTag(new IFDEntry(Tag.YResolution, FieldType.RATIONAL).setValue(1, 1));
       geotiff.addTag(new IFDEntry(Tag.ResolutionUnit, FieldType.SHORT).setValue(1));
-
+      // black is zero (value used in GeotiffWriter)
+      geotiff.addTag(new IFDEntry(Tag.PhotometricInterpretation, FieldType.SHORT).setValue(1));
     } else {
+      if (colorTable != null && colorTable.length > 0) {
+        // standard tags for Palette-color images ( see TIFF spec, section 5)
+        geotiff.addTag(new IFDEntry(Tag.PhotometricInterpretation, FieldType.SHORT).setValue(3));
+        geotiff.addTag(new IFDEntry(Tag.ColorMap, FieldType.SHORT, colorTable.length).setValue(colorTable));
```
WeatherGod: Just realized that I need to have this in the greyscale mode, too, if I want to be consistent with what I documented. I should also add a unit test proving out that this works.

WeatherGod: Just fixed this, and I think I vastly simplified `writeMetadata()`, too.

WeatherGod:
Test failures seem to be unrelated?

tdrwenski:
> Test failures seem to be unrelated?

Hi @WeatherGod, I think I know the issue there, should have a fix shortly!

WeatherGod:
Is there anyplace where I should add some narrative docs explaining the changes to the geotiff module?

tdrwenski:
@WeatherGod, sorry for the delay getting back to this. We don't have any documentation that I can find about geotiffs. So I think the best spot for documentation is in the javadocs for these classes. We can also add a line to our next release's notes like "Support for saving Geotiffs with a palette", that will hopefully alert users to have a look at your changes.

WeatherGod:
Ok, that makes sense. Do you want something more than what I've written so far?

tdrwenski (approved):

tdrwenski:
Looks good to me! Thanks!

tdrwenski:
Failing test is a flaky test unrelated to your changes and we are fixing it in a separate PR

WeatherGod:
Thank you very much! Question: does the CI/CD for the tds repo use the latest release of netcdf-java or the development branch? I have what is essentially the second half of this feature which allows users to download geotiffs with the colormap from thredds WCS service. It isn't as fully fleshed out as this one was, but it would be nice to at least get a draft PR up.

haileyajohnson:
> Thank you very much! Question: does the CI/CD for the tds repo use the latest release of netcdf-java or the development branch? 

It uses the latest netcdf-java snapshot, which is functionally the development branch.

PR #1071, opened 2022-08-26
[5.x] Improve error message

tdrwenski:
## Description of Changes

A user had a netCDF file with a time unit set to "Year". I learned that this is apparently not allowed in the [CF convention](https://cfconventions.org/cf-conventions/cf-conventions.html#time-coordinate) since you must have a reference time (for example "Year since 1970").

However, this lead to a time axis that had the name "time" but that was not used as a coordinate. So when the variable that depended on "time lon lat" was requested through NCSS, the `DtCoverage::readDataSection` threw
```
Throwable exception handled : java.lang.ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2
```
because the latitude was index 2 for the variable but the subset requested did not include time (had dimension 2). Which isn't a very user friendly message.

This PR improves the error message the user gets when requesting the variable to:
```
Throwable exception handled : ucar.ma2.InvalidRangeException: Variable "myVariableName" has different number of dimensions than the coordinate system. This is likely due to a problem with a dimension in the data.
 Variable dimensions: time longitude latitude.
 Coordinate system: Coordinate System (time latitude longitude)
 rt=latitude(latitude=720) (ucar.nc2.dataset.CoordinateAxis1D)
 rt=longitude(longitude=1440) (ucar.nc2.dataset.CoordinateAxis1D)
 Projection: LatLonProjection Center lon:.0
 ```
 
 as well as logging when an axis is invalid due to bad units.

## PR Checklist

       until ready for review

### CODE REVIEW

File: cdm/core/src/main/java/ucar/nc2/internal/dataset/conv/DefaultConventions.java [hunk 1]
```diff
@@ -213,12 +213,14 @@ protected AxisType getAxisType(VariableDS.Builder vb) {
       {
         return AxisType.Height;
       }
+      logger.info("Unit '" + unit + "' is not compatible with meters for axis '" + vname + "'");
```
tdrwenski, line 216: This class also has a `parseInfo` I could log to but I am not sure where those messages end up

haileyajohnson (approved):

Issue #1060, opened 2022-08-11
[5.5.3]: Fix GRIB IOSP to ensure time coordinates are monotonically increasing


ethanrd:
### Versions impacted by the bug

v5.x

### What went wrong?

For variables where the data values are valid on a time interval, e.g., total precipitation, the GRIB IOSP calculates time coordinates from the set of time intervals specified in the GRIB records. When the set of time intervals are irregular and overlapping, using the mid-point of each interval as the time coordinate values can result in a coordinate variable that is not monotonically increasing.

### If you have an example file that you can share, please attach it to this issue.
If so, may we include it in our test datasets to help ensure the bug does not return once fixed?
Note: the test datasets are publicly accessible without restriction.

N/A

### Code of Conduct

- [X] I agree to follow the UCAR/Unidata Code of Conduct

ethanrd:
This issue (and soon to be added PRs) replaces PR #1043 in order to separate the solution for datasets with 1-D time from the solution for FMRC datasets with 2-D time (model run and valid).

ethanrd:
I have solution for datasets with 1-D time and will submit soon, probably early next week. Still working on solution for FMRC datasets with 2-D time.

pedro-cf:
NetCDF example: https://drive.google.com/file/d/1ypnvsHarmJxZgeueGXXSDw4yVNaSi_mO/view?usp=sharing

pedro-cf:
> NetCDF example: https://drive.google.com/file/d/1ypnvsHarmJxZgeueGXXSDw4yVNaSi_mO/view?usp=sharing

this same NetCDF can be accessed via this thredds server (wms): http://thredds.maretec.org/thredds/catalog/IST_MOHID_BIO_DATA/TAGUS_0.007DEG_50L_1H/FORECAST/catalog.html?dataset=MOHIDWATERDatasetScan/TAGUS_0.007DEG_50L_1H/FORECAST/2023051700.nc

PR #1072, opened 2022-08-26
[5.x]: Fix GRIB IOSP so time coords are monotonically increasing, for 1-D time datasets only

ethanrd:
## Description of Changes

Fix GRIB IOSP code that creates time coordinate variables from a set of irregularly overlapping time intervals to ensure the resulting time coordinate is monotonically increasing. This fix only applies to datasets with 1-D time (not FMRCs with 2-D time).

Rather than simply using the interval mid-points, start with the mid-points and then use the current and previous time intervals to ensure the resulting time coordinate is monotonically increasing.

This change includes test and test dataset. Test passes with this change.

## PR Checklist
       until ready for review


ethanrd:
This PR addresses part of issue #1060; it addresses datasets with 1-D time. (PR for 2-D time datasets still to come.)

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private void checkTimeCoordinateVariable1D_IsMonotonicallyIncreasing(int timeDimLength, Array timeCoordValues) {
+    double currentValue = timeCoordValues.getDouble(0);
+    double prevValue = currentValue;
+    StringBuilder valuesSoFar = new StringBuilder();
+    valuesSoFar.append(currentValue);
+    for (int i = 1; i < timeDimLength; i++) {
+      currentValue = timeCoordValues.getDouble(i);
+      valuesSoFar.append(", ").append(currentValue);
+      Assert.assertTrue("Not monotonically increasing: [" + valuesSoFar + "]", currentValue > prevValue);
```
tdrwenski: for monotonic, shouldn't it be equality be allowed, I mean like:
```suggestion
      Assert.assertTrue("Not monotonically increasing: [" + valuesSoFar + "]", currentValue >= prevValue);
```

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
```diff
@@ -0,0 +1,50 @@
...
+import ucar.ma2.Array;
+import ucar.nc2.*;
+
+import java.io.IOException;
+
+public class TestDiscontiguousInterval {
+
+  @Test
+  public void testTimeCoordinate1D_isMonotonicallyIncreasing() throws IOException {
+    String testfile = "../grib/src/test/data/GFS_Global_onedeg_20220627_0000.precipOnly.grib2";
```
tdrwenski: Is it possible to make a smaller test file? I think 28 MB is a bit on the large size. Otherwise we could use an existing one in the `thredds-test-data` repo

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
```diff
@@ -787,11 +787,26 @@ private Array makeLazyTime2Darray(Variable coord, Time2Dinfo info) {
         break;
 
       case intvU:
```
tdrwenski, line 787: For my understanding, what does this mean again? interval...  unique? 😅

haileyajohnson (approved):

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
```diff
@@ -787,11 +787,26 @@ private Array makeLazyTime2Darray(Variable coord, Time2Dinfo info) {
...
+          TimeCoordIntvValue prevTimeIntervalValue = null;
           for (TimeCoordIntvValue tinv : timeIntv.getTimeIntervals()) {
-            data[count++] = timeUnit.getValue() * tinv.getCoordValue() + time2D.getOffset(runIdx);
+            data[count] = timeUnit.getValue() * tinv.getCoordValue() + time2D.getOffset(runIdx);
+            if (count >= 2) {
+              // Check that coordinate values are monotonically increasing
+              if (data[count] <= data[count - 1]) {
+                // Set value to half way between previous coord value and current interval upper bound.
+                // Unless those are the same, then add the distance between the previous
+                // and current upper bounds to the ???? Ugh. Hard to describe in words. Will think more.
```
tdrwenski: update comment here?

tdrwenski: Also, does equality of the neighboring coordinates need to be "fixed" to be monotonic? I thought monotonic just meant x <= y means f(x) <= f(y)

edit: duplicate of above discussion on monotonic vs strictly monotonic

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
```diff
@@ -787,11 +787,26 @@ private Array makeLazyTime2Darray(Variable coord, Time2Dinfo info) {
...
+                if (tinv.getBounds2() != prevTimeIntervalValue.getBounds2())
+                  data[count] = (tinv.getBounds2() - prevTimeIntervalValue.getBounds2()) / 2.0 + data[count - 1];
```
tdrwenski: maybe you need to take the units and offsets into account? like we do on line 795 above

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
ethanrd: Interesting. The [NUG defines "coordinate variable" including this phrase] (https://docs.unidata.ucar.edu/nug/current/netcdf_data_set_components.html#coordinate_variables) "strictly monotonic (all values are different and either increasing or decreasing)". While [CF says](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#terminology) it uses the term "coordinate variable" precisely as defined in the NUG but then drops "strictly" in the text.

I believe coordinate variables need to be "strictly monotonic" to support transformations between coordinate systems and the like. So I think the ">" is correct. I will add "strictly" to the language in this PR to clarify. I also think I will open issues on CF to update the language so that it is clear that values must be unique.

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
ethanrd: Yup. I will look at how small I can get it.

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
ethanrd: I think so. I'm just not quite sure I understand what "unique" means.

Also, the test dataset has 1D time coordinates but methods and such in code include 2D ,e.g., "[makeLazyTime2Darray(...)](https://github.com/Unidata/netcdf-java/blob/2f872bd32b33a662a66bcb00707239b28632be02/grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java#L737)" and "enum GribIosp.[Time2DinfoType](https://github.com/Unidata/netcdf-java/blob/2f872bd32b33a662a66bcb00707239b28632be02/grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java#L568)" (which includes an "is1Dtime")?

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
ethanrd: Right. I will update that comment.

Same as above, going with "strictly monotonic".

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
ethanrd: Yup. Thanks! I'll get that fixed.

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
tdrwenski: This same function exists in another class, the `GribIospBuilder` [here](https://github.com/Unidata/netcdf-java/blob/2f872bd32b33a662a66bcb00707239b28632be02/grib/src/main/java/ucar/nc2/grib/collection/GribIospBuilder.java#L537). It appears to be unused, so I don't think you need to update it. I will look into if I can remove it.

ethanrd: Yup, I saw that too. Didn't poke around enough to figure out if it was ever used. Looks like there is code in GribIosp that does instantiate a GribIospBuilder and then work with it but I haven't dug into it to see what cases might actually use the builder.

I did look at netcdf-java-ng and lots of those methods are removed from GribIosp and remain in GribIospBuilder. So, I think 5.x got the start but not completion of the transition to builders. I was hoping netcdf-java-ng might also clarify the difference between `makeLazyTime1Darray()` and `makeLazyTime2Darray()`. Unfortunately both still have the same "// only for 2d times" comment found in 5.x.

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
ethanrd: I think I addressed this with latest commit. So will mark resolved.

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
ethanrd: Switched to a much smaller test dataset with latest commit. So will mark resolved.

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
ethanrd: Finished, clarified, and cleaned up comments in latest commit. So will mark resolved.

grib/src/main/java/ucar/nc2/grib/collection/GribIosp.java
ethanrd: Addressed this with latest commit. So will mark resolved.

tdrwenski (approved):

ethanrd:
Thanks Tara @tdrwenski !

PR #1075, opened 2022-08-29
Cleanup GribIosp

tdrwenski:
## Description of Changes

 - Remove unused private methods (these seem to have been accidentally copy/pasted from GribIosp.java)
 - Remove unused import
 - Fix typos
 - Compare strings with equals not !=
 - Remove unused variable

## PR Checklist

- [x] Add labels
       until ready for review


ethanrd:
Huh. While working on my strictly monotonic PR I was wondering why I wasn't seeing any use of `GribIospBuilder` code. Still not seeing why it isn't used when opening a dataset with the new `NetcdfFiles` (with an "s") API which seems set up to use the builder. Of course, I haven't stepped through that code in a debugger so probably just missed something obvious while browsing the code.

tdrwenski:
@ethanrd, interesting (also saw your comment on your PR about it!). Now I looked at version 8 and I see `makeLazyCoordinateArray` is now a static helper in the class `Time2DLazyCoordinate`. And the one I removed here, the `GribIospBuilder::makeLazyCoordinateData` is still unused. 

I think like you said, the `NetcdfFiles.open` uses the `GribIospBuilder` via the `GribIosp::build` method. However, the `makeLazyCoordinateData` functions only seem to be called from `GribIosp::readArrayData` method, and they act on a `Variable` which may also only be available when you are reading the data and not when you are building. I guess that reading data happens after `build` and there is no `Builder` object available at that time. So that was why I was thinking these methods should be removed and can't actually be used in the `Builder`.  If that makes sense.

ethanrd:
Ah, yes. The lazy eval doesn't eval until after the Builder is built. I didn't quite make that connection. Doh!

ethanrd (approved):

PR #1080, opened 2022-09-08
Handle HDF5 file when a variable has a zero dimension

### LINKED ISSUE #1037: [VERSION]: 5.5.2 Dimension length =0 must be > 0

v5.x

### What went wrong?

Unable to open hdf5 file which fails with error: `Dimension length =0 must be > 0`. 

The same file works fine when using the deprecated `NetcdfFile.open(...)` instead of newer `NetcdfFiles.open(...)`. The deprecated function, logs error but continues to read the file successfully. Error Logged:
`DEBUG ucar.nc2.NetcdfFile - Using IOSP ucar.nc2.iosp.hdf5.H5iosp
ERROR ucar.nc2.iosp.hdf5.H5header - shape[0]=0 must be > 0
DEBUG ucar.nc2.iosp.hdf5.H5header - ERROR: makeVariableShapeAndType shape[0]=0 must be > 0` 

Given our scenario, we are not processing the variables that causes the issue, hence we would like to continue reading the file ignoring the error. The hdf5 files we use for processing are from our external partners. 

Is it possible to have the same behavior of ignoring the error be added to new implementation as well? It comes down to the H5Header.java -> makeVariableShapeAndType -> which has a try/catch that handles the InvalidRangeException whereas in new implementation, its thrown as Runtime - IllegalArgumentException. 

### Relevant stack trace

```shell
Caused by: java.io.IOException: java.lang.IllegalArgumentException: 
	at ucar.nc2.NetcdfFiles.open(NetcdfFiles.java:279)
	at ucar.nc2.NetcdfFiles.open(NetcdfFiles.java:243)
	at ucar.nc2.NetcdfFiles.open(NetcdfFiles.java:216)
	... 33 more
Caused by: java.lang.IllegalArgumentException: Dimension length =0 must be > 0
	at ucar.nc2.Dimension$Builder.setLength(Dimension.java:582)
	at ucar.nc2.Dimensions.makeDimensionsAnon(Dimensions.java:141)
	at ucar.nc2.Variable$Builder.setDimensionsAnonymous(Variable.java:2061)
	at ucar.nc2.internal.iosp.hdf5.H5headerNew.makeVariableShapeAndType(H5headerNew.java:1565)
	at ucar.nc2.internal.iosp.hdf5.H5headerNew.makeVariable(H5headerNew.java:1326)
	at ucar.nc2.internal.iosp.hdf5.H5headerNew.makeNetcdfGroup(H5headerNew.java:549)
	at ucar.nc2.internal.iosp.hdf5.H5headerNew.makeNetcdfGroup(H5headerNew.java:539)
	at ucar.nc2.internal.iosp.hdf5.H5headerNew.makeNetcdfGroup(H5headerNew.java:539)
	at ucar.nc2.internal.iosp.hdf5.H5headerNew.makeNetcdfGroup(H5headerNew.java:539)
	at ucar.nc2.internal.iosp.hdf5.H5headerNew.makeNetcdfGroup(H5headerNew.java:539)
	at ucar.nc2.internal.iosp.hdf5.H5headerNew.read(H5headerNew.java:264)
	at ucar.nc2.internal.iosp.hdf5.H5iospNew.build(H5iospNew.java:122)
	at ucar.nc2.NetcdfFiles.build(NetcdfFiles.java:811)
	at ucar.nc2.NetcdfFiles.open(NetcdfFiles.java:750)
	at ucar.nc2.NetcdfFiles.open(NetcdfFiles.java:276)
	... 36 more
```

### Relevant log messages

_No response_

No

### Code of Conduct

mnlerman:
## Description of Changes

_Mimic behavior of version 4 when reading a HDF5 file with a variable where one of the dimensions is 0. Provide warning in the log but do not error out. Addresses Unidata/netcdf-java/issues/1037_

## PR Checklist

- [x] Add labels
       until ready for review

### DIFF
--- a/bufr/src/main/java/ucar/nc2/iosp/bufr/BufrIospBuilder.java
+++ b/bufr/src/main/java/ucar/nc2/iosp/bufr/BufrIospBuilder.java
@@ -124,13 +124,12 @@ class BufrIospBuilder {
     dkey.name = uname; // name may need to be changed for uniqueness
 
     Structure.Builder struct = Structure.builder().setName(uname);
-    try{
+    try {
       struct.setDimensionsAnonymous(new int[] {count}); // anon vector
-    }
-    catch (InvalidRangeException e){
+    } catch (InvalidRangeException e) {
       log.error(e.getMessage());
     }
-      for (BufrConfig.FieldConverter subKey : fld.flds) {
+    for (BufrConfig.FieldConverter subKey : fld.flds) {
       addMember(group, struct, subKey);
     }
 
@@ -182,10 +181,9 @@ class BufrIospBuilder {
     Structure.Builder struct = Structure.builder().setName(uname);
     parent.addMemberVariable(struct);
     int n = parentFld.dds.replication;
-    try{
+    try {
       struct.setDimensionsAnonymous(new int[] {n}); // anon vector
-    }
-    catch (InvalidRangeException e){
+    } catch (InvalidRangeException e) {
       log.error(e.getMessage());
     }
 
@@ -202,10 +200,9 @@ class BufrIospBuilder {
 
   private void addDpiSequence(Structure.Builder parent, BufrConfig.FieldConverter fld) {
     Structure.Builder struct = Structure.builder().setName("statistics");
-    try{
+    try {
       struct.setDimensionsAnonymous(new int[] {fld.dds.replication}); // scalar
-    }
-    catch (InvalidRangeException e){
+    } catch (InvalidRangeException e) {
       log.error(e.getMessage());
     }
 
@@ -228,10 +225,9 @@ class BufrIospBuilder {
 
     Variable.Builder v = Variable.builder().setName(uname);
     if (count > 1) {
-      try{
+      try {
         v.setDimensionsAnonymous(new int[] {count}); // anon vector
-      }
-      catch (InvalidRangeException e){
+      } catch (InvalidRangeException e) {
         log.error(e.getMessage());
         log.debug("ERROR: makeVariableShapeAndType {}", e.getMessage());
       }
@@ -260,10 +256,9 @@ class BufrIospBuilder {
     if (dataDesc.type == 1) {
       v.setDataType(DataType.CHAR);
       int size = dataDesc.bitWidth / 8;
-      try{
+      try {
         v.setDimensionsAnonymous(new int[] {size});
-      }
-      catch (InvalidRangeException e){
+      } catch (InvalidRangeException e) {
         log.error(e.getMessage());
         log.debug("ERROR: makeVariableShapeAndType {}", e.getMessage());
       }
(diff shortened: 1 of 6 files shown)


bufr/src/main/java/ucar/nc2/iosp/bufr/BufrIospBuilder.java
```diff
@@ -123,7 +124,11 @@ private void addStructure(Group.Builder group, Structure.Builder parent, BufrCon
     dkey.name = uname; // name may need to be changed for uniqueness
```
haileyajohnson, line 125: I don't know that this method of catching will work out as well in bufr, especially with structs. I would lean toward letting bufr throw throw the error if the dimension is 0, since the issue only complained about hdf5 datafiles.

cdm/core/src/main/java/ucar/nc2/Variable.java
```diff
@@ -2055,9 +2055,13 @@ public String makeDimensionsString() {
      * Set the dimensions using all anonymous (unshared) dimensions
      *
      * @param shape defines the dimension lengths. must be > 0, or -1 for VLEN
-     * @throws RuntimeException if any shape < 1 and not -1.
+     * @throws InvalidRangeException if any shape < 1 and not -1.
      */
-    public T setDimensionsAnonymous(int[] shape) {
+    public T setDimensionsAnonymous(int[] shape) throws InvalidRangeException {
```
tdrwenski, line 2060: The `InvalidRangeException` definitely makes more sense than throwing for instance a `RuntimeException`, but I think changing this function's signature could be considered a breaking change-- since this function is `public` and technically anyone could use it in their code.

mnlerman: Ahh yeah....that does make sense. Do you suggest removing the method signature and just throwing the exception? To be fully honest, I'm still trying to get oriented and copying the conventions I've seen elsewhere in the code without knowing the full implications of the changes.

haileyajohnson: I think the only way to accomplish this without changing the method signature is to return `null` (or something) instead of throwing the exception, then in places where you have the `try/catch`, check for `null` instead.

tdrwenski: Or instead of throwing the `InvalidRangeException` here, you could catch and log `IllegalArgumentException`s. Those come from the `Dimension.makeDimensionsAnon` which calls `Dimension::setLength`. I think that does a similar check on the shape that you are doing here.

I guess based on Hailey's comment (that we only want to change the behavior for hdf5 and not bufr) that it is best to do the catching and logging in the `H5headerNew` where this is called (like you are currently doing for the `InvalidRangeException`)

mnlerman:
Okay, I _think_ I have a better solution but I have no idea how to test it. I'm going to close this PR and create a new one

Issue #126, opened 2019-09-21
dap4.test.TestNc4Iosp failure on Jenkins


lesserwhirls:
We have a new failure on Jenkins for `dap4.test.TestNc4Iosp.testNc4Iosp`. This started showing up when the way we load IOSPs changed with PR 101 (see https://github.com/Unidata/netcdf-java/pull/101/files for the changes).

My suspicion is that the change caused the dap4 library to use the `Hdf5Iosp` instead of the `Nc4Iosp`. The code should handle both cases, but I think what we see is that there is a difference in the way the way the two IOSPs see variable metadata (in this case, something about the way `enum` is handled). Here is the output from Jenkins:

~~~bash
Netcdf-c library version: 4.6.1 of Mar 30 2018 02:30:19 $
Testcase: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/testfiles/test_one_var.nc
Testpath: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/testfiles/test_one_var.nc
Baseline: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/TestIosp/baseline/test_one_var.nc.nc4
DMR Comparison:
Files are Identical
DATA Comparison:
Files are Identical
Testcase: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/testfiles/test_one_vararray.nc
Testpath: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/testfiles/test_one_vararray.nc
Baseline: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/TestIosp/baseline/test_one_vararray.nc.nc4
DMR Comparison:
Files are Identical
DATA Comparison:
Files are Identical
Testcase: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/testfiles/test_atomic_types.nc
Testpath: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/testfiles/test_atomic_types.nc
Baseline: /home/ubuntu/jenkins/workspace/netcdf-java/dap4/d4tests/src/test/data/resources/TestIosp/baseline/test_atomic_types.nc.nc4
DMR Comparison:
[...]

PR #1091, opened 2022-09-30
Fix EnumTypedef name problem

DennisHeimbigner:
## Description of Changes
re: Issue https://github.com/Unidata/netcdf-java/issues/126

The code for handling enum types in Group.java is
incorrect. When creating a new enum type, it is appropriate to
search only the current group for a conflicting name and this is
what the current code does.  But when an enum typed variable
searches for the matching enum type, it must search not only the
current group but all parent groups and the current code does
not do that.

The fix consists of two similar parts.
1. Modify Group.findEnumeration to take an extra boolean parameter
to control if search is this group only or to search up the group's parents.
2. Modify Group.Builder.findEnumTypedef to act like part 1 but to search
the sequence of parent Group.Builder instances.

As a consequence, this PR modifies a number of other files to
adapt to the modified signatures.

## Note
This same problem appears to affect the opaque type also, but
fixing that may be a bit more difficult because CDM appears to
convert opaque types to byte typess.

## PR Checklist

- [ ] Add labels
       until ready for review

## Addendum 1
It turns out that I missed the error in the code in H5headerNew that
attempts to match an enum typed variable with the proper enumeration
type declaration.

The problem and fix (as described by a comment in H5headerNew)
is a bit of a hack. It should be fixed in H5Object.read().
Unfortunately, information is not being passed down so that
the proper fix can be applied early in the HDF5->CDM translation.
Fixing this would affect a lot of function signatures.

Also modified TestEnumTypedef.java to test two cases:
1. the actual enum type def is in the same group as the variable that uses it.
2. the actual enum type def is in a parent group of the variable that uses it

## Misc. Other Changes
* Suppress printing of _NCProperties to simplify text-based comparison testing.

## Addendum 2 (12/19/2022)
* NetcdfFile.java: convert to use H5iospNew (needed by TestH5iosp.java)
* H5headerNew.java: provide get function for accessing the btree (needed by TestDataBTree.java)
* H5iospNew.java: make getRandomAccessFile() method public (needed by tests)
* CompareNetcdf2.java: Add a constructor to specify if attribute name comparison should ignore case or not. It turns out that some tests require case sensitive name matching. Specifically TestCoordSysCompare.java and TestN3iospCompare.java

## Addendum 3 (2/25/2023)
Make additional changes to fix Jenkins failures.

* Revert Group.attributes()
* Revert the use of H5iospNew in NetcdfFile.
* Add to Group a method that looks for a matching enumeration purely structurally, as opposed to name matching. This corresponds to the same method in Group Builder.
* Rebuild the EnumTypedef search algorithm in H5headerNew and H5header to handle cases in Jenins only enum tests.
* Revert various calls to Group.findEnumeration().
* Add some test case data: ref_anon_enum.h5 and test_enum_type.nc
* Convert TestEnumTypedef to JUNIT.parameter form.
* Revert TestH5iosp.
* The URL used in TestSequence is no longer valid, so change to an equivalent url on the remotetest server.


cdm/core/src/main/java/ucar/nc2/Group.java
```diff
@@ -376,18 +377,20 @@ public boolean removeAttributeIgnoreCase(String attName) {
 
   /** Find a Enumeration in this or a parent Group, using its short name. */
   @Nullable
-  public EnumTypedef findEnumeration(String name) {
+  public EnumTypedef findEnumeration(String name, boolean searchup) {
```
haileyajohnson, line 388: Since this is public, we can't change the method signature. We can add a second method signature though.

cdm/core/src/main/java/ucar/nc2/Group.java
```diff
@@ -1114,6 +1117,26 @@ public Group.Builder commonParent(Group.Builder other) {
       return other;
     }
 
+    /** Find a Enumeration in this or a parent Group Builder, using its short name. */
+    public Optional<EnumTypedef> findEnumTypedef(String name, boolean searchup) {
```
haileyajohnson, line 1155: Same here, we'll need one with a boolean parameter and one with a default.

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
```diff
@@ -44,11 +39,11 @@
 import ucar.nc2.Structure;
 import ucar.nc2.Variable;
 import ucar.nc2.constants.CDM;
-import ucar.nc2.filter.Filter;
 import ucar.nc2.internal.iosp.hdf4.HdfEos;
 import ucar.nc2.internal.iosp.hdf4.HdfHeaderIF;
 import ucar.nc2.internal.iosp.hdf5.H5objects.DataObject;
 import ucar.nc2.internal.iosp.hdf5.H5objects.DataObjectFacade;
+import ucar.nc2.internal.iosp.hdf5.H5objects.Filter;
 import ucar.nc2.internal.iosp.hdf5.H5objects.GlobalHeap;
```
haileyajohnson: Is this being used?

dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
```diff
@@ -0,0 +1,406 @@
...
+/**
+ * Test OpenDap Server at the NetcdfDataset level
+ */
+public class TestDAP4Client extends DapTestCommon {
+  private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  static final boolean DEBUG = false;
+
+  static final boolean BROKEN = false; // on/off known broken tests
+
```
haileyajohnson, line 28: We can use annotations to turn off known broken tests, we shouldn't need a flag.

dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
```diff
@@ -0,0 +1,406 @@
...
+      chosentests = locate("test_atomic_types.nc");
+      prop_baseline = true;
+    } else {
+      for (ClientTest tc : alltestcases) {
+        chosentests.add(tc);
+      }
+    }
+  }
+
+  boolean defineAllTestcases(String root, String servlet, String server) {
```
haileyajohnson, line 222: JUnit parameterized?

haileyajohnson (changes requested):

dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
haileyajohnson: I'm happy to take a pass at converting this test to use parameterized if you'd like

dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
DennisHeimbigner: Do you object to overloading the method name or should I use
a different name?

cdm/core/src/main/java/ucar/nc2/Group.java
DennisHeimbigner: Do you object to overloading the method name or should I use
a different name?

dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
DennisHeimbigner: This will be fixed in a later PR addressing major changes to DAP4

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
DennisHeimbigner: Is there a gradle task for checking for unused imports?

cdm/core/src/main/java/ucar/nc2/Group.java
haileyajohnson: Overloading is fine

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
haileyajohnson: There may be one, but we don't have it implemented, unfortunately.

dap4/d4tests/src/test/java/dap4/test/TestDAP4Client.java
DennisHeimbigner: I would say don't bother with fixing DAP4 tests at the moment.
This PR addresses issues completely independent of DAP4.
I do not think there is any point in trying to add DAP4 fixes to it
since I have another branch that is working on that. In any case,
converting the DAP4  tests to use parameterized tests is going to take
time. I think we will have to live with it for a while.

haileyajohnson (approved):

DennisHeimbigner:
I give up. The failing test is apparently referencing the file
> /share/testdata/cdmUnitTest/thredds-test-data/local/thredds-test-data/cdmUnitTest/formats/hdf5/samples

and I have no idea how to access it to see what is going on.

tdrwenski:
Hi Dennis, the test data in that folder comes from here: https://github.com/Unidata/thredds-test-data. There is a ton of data which takes very long to download. I can try to send or point you to the relevant subset of the data or you can run the tests on Jenkins against your branch.

Issue #1045, opened 2022-07-07
[5.5.3+]: Failure to build netcdfAll with optional zarr


rschmunk:
### Versions impacted by the bug

v5.x

### What went wrong?

A user developing a data product asked whether it would be possible to modify Panoply to access data in zarr directories or in zarr zip files. After looking through documentation, it looked like the first step was to build a netcdfAll where the config is modified to include `cdm:zarr`.

Using a netcdf-5.5.3 SNAPSHOT from January 2022, I succeeded in doing so by adding

`netcdfAll project(':cdm:cdm-zarr')
`

at line 54 of `fatJars.gradle`. And when using the resulting netcdfAll jar I have been able to successfully open an example zarr zip file.

However, if I try to start from a later netcdf-5.5.3 SNAPSHOT (e.g., one from April), or the 5.5.3 release, or the current 5.5.4 SNAPSHOT, the build fails with a complaint about missing dependencies. As follows...

```
* What went wrong:
Could not determine the dependencies of task ':buildNetcdfAll'.
> Could not resolve all dependencies for configuration ':netcdfAll'.
   > Could not find com.fasterxml.jackson.core:jackson-core:.
     Required by:
         project : > project :cdm:cdm-zarr
   > Could not find com.fasterxml.jackson.core:jackson-databind:.
     Required by:
         project : > project :cdm:cdm-zarr
```

Is this due to a bug in the gradle config for `cdm:zarr`? Is there something else in the config that I need to edit in order to include `cdm:zarr` in the netcdfAll build? My knowledge of gradle remains sketchy enough that I wouldn't be surprised if it is the latter.

Also, while looking into this, I ran into the issue that `ZarrHeader` does not bother to apply dimension names that are provided via the `_ARRAY_DIMENSIONS` variable attribute. I'll try to upload a pull request to fix that soon, but I'd like to figure out the build problem first.

### If you have an example file that you can share, please attach it to this issue.
If so, may we include it in our test datasets to help ensure the bug does not return once fixed?
Note: the test datasets are publicly accessible without restriction.

No

### Code of Conduct

- [X] I agree to follow the UCAR/Unidata Code of Conduct

haileyajohnson:
@rschmunk really glad people are starting to try out the Zarr package! 
I can reproduce the error you're getting, it seems to be related to transitive dependencies (or loss of) from upgrading the AWS SDK, and I'm troubleshooting it now.

[...]

PR #1098, opened 2022-10-14
clean up zarr build and check for null location

haileyajohnson:
- remove unwanted dependencies from zarr module
- remove jackson-bom (netcdfAll can't get versions for core and databind from it?)
- handle null location in MZipFileProvider

### DIFF
--- a/netcdf-java-platform/build.gradle
+++ b/netcdf-java-platform/build.gradle
@@ -10,7 +10,9 @@ javaPlatform {
 
 dependencies {
   def awsVersion = '2.17.290'
+  def jacksonVersion = '2.14.0-rc1'
   api enforcedPlatform("software.amazon.awssdk:bom:${awsVersion}")
+  api enforcedPlatform("com.fasterxml.jackson:jackson-bom:${jacksonVersion}")
   constraints {
     // Note: The depVersion variable is defined in gradle/any/shared-mvn-coords.gradle and is used for dependencies
     // that are used in different configurations of a build that need access to the full maven coordinates.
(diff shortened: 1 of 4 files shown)


netcdf-java-platform/build.gradle
```diff
@@ -11,7 +11,6 @@ javaPlatform {
 dependencies {
   def awsVersion = '2.17.290'
   api enforcedPlatform("software.amazon.awssdk:bom:${awsVersion}")
-  api enforcedPlatform('com.fasterxml.jackson:jackson-bom:2.14.0-rc1')
```
lesserwhirls, line 14: I would keep the `jackson-bom`, as it can control the transitive dependency versions (as well as making sure other jackson artifacts not used directly by netcdf-java are the same version as the ones we use). In order to have `netcdfAll` find the dependencies, you'll need to keep the the `jackson-core` and `jackson-databind` api lines below, and define a `jacksonVersion` variable like I did with `awsVersion` for those two lines to reference. Something like:

```
  def awsVersion = '2.17.290'
  def jacksonVersion = '2.14.0-rc1'
  api enforcedPlatform("software.amazon.awssdk:bom:${awsVersion}")
  api enforcedPlatform("com.fasterxml.jackson:jackson-bom:${jacksonVersion}")
  constraints {
    ...
    // netcdfAll (everything else is happy with the jackson BOM)
    api "com.fasterxml.jackson.core:jackson-core:${jacksonVersion}"
    api "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"
    ...
```

haileyajohnson: We're not actually even putting zarr in netcdfAll but #1045 :(

haileyajohnson: also do you know _why_ it doesn't work for netcdfAll?

lesserwhirls: I think it does not work because we try to get away with not having essentially empty projects for netcdfAll, toolsUI, and ncIdv (a folder and single build.gradle for each one). We create custom configurations for those projects and attach dependencies to those configurations, and that does not seem to be able to work with dependencies that are BOMs within platform projects. I bet if you created actual subprojects and declared dependencies using implementation or runtimeOnly, it would work as expected (but then you'd have more directories cluttering up the layout).

tdrwenski (approved):

rschmunk:
I just downloaded the latest 5.5.4 SNAPSHOT and successfully built a netcdfAll with Zarr included, so this apparently fixes #1045.

I was also able to make a 2D plot of uncompressed Zarr data, which I hadn't been able to do before. So I guess there are other Zarr bug fixes since the NJ 5.5.3 SNAPSHOT from January that I had been using to test Zarr access. I'll have to explore some more with the handful of Zarr samples I have.

PR #1099, opened 2022-10-14
Fix grib config time unit conversion

tdrwenski:
## Description of Changes

Sometimes, like on our thredds server, a grib collection has some files/records with time units in minutes, and some in hours. Then variables that depend on that time coordinate will be labeled "mixed_intervals" even if the intervals are the same lengths but with different units. This PR addresses issues with unit conversion so that we can fix this issue on our thredds server by adding something like this to a grib collection we want to have units in hours:

```
    <gribConfig>
      <timeUnitConvert from="0" to="1"/>
    </gribConfig>
```

- Add tests for various changes. Also added two new files to the thredds-test-data on lead for these tests.
- Fix rounding of time interval offsets in `CalendarPeriod`: previously interval offsets (with respect to a reference) were rounded towards zero, which caused the intervals to be shorter if the reference date came after the start than if it came before the start.
- Fix bug in the logic of unit conversion in `getForecastTimeInterval`
- Allow `setTimeUnitConverter` to be modified after it is set. If you try to update a collection index, then it may already have a `timeUnitConverter` set and try to reset it, so will fail to update the index. Maybe users shouldn't call this and expect it to change the units after the index is created though, so I kept a comment here.
- I have not updated the documentation that says not to use the parameter `timeUnitConvert`. Perhaps we should first make sure this fix is working as we expect and that more cases are tested before telling users to use this.

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Tables.java
+++ b/grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Tables.java
@@ -564,7 +564,7 @@ public class Grib2Tables implements ucar.nc2.grib.GribTables, TimeUnitConverter 
     if (wantPeriod == null) {
       return null;
     }
-    CalendarPeriod havePeriod = Grib2Utils.getCalendarPeriod(convertTimeUnit(intvu.timeUnitIntv));
+    CalendarPeriod havePeriod = Grib2Utils.getCalendarPeriod(intvu.timeUnitIntv);
     double fac2 = intvu.timeRange * wantPeriod.getConvertFactor(havePeriod);
     CalendarPeriod period = wantPeriod.multiply((int) fac2); // LOOK needs to be an int
 
--- a/grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Tables.java
+++ b/grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Tables.java
@@ -38,6 +38,7 @@ import java.util.*;
  * @since 4/3/11
  */
 @Immutable
+// TODO version 6: remove implements TimeUnitConverter
 public class Grib2Tables implements ucar.nc2.grib.GribTables, TimeUnitConverter {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Grib2Tables.class);
   private static final Map<Grib2TablesId, Grib2Tables> tables = new HashMap<>();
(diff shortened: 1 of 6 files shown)


grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Tables.java
```diff
@@ -460,9 +460,8 @@ public String getLevelNameShort(int id) {
 
   private TimeUnitConverter timeUnitConverter; // LOOK not really immutable
 
+  // Do not modify timeUnitConverter once it's been set
   public void setTimeUnitConverter(TimeUnitConverter timeUnitConverter) {
-    if (this.timeUnitConverter != null)
```
haileyajohnson, line 464: Is it ok to remove this?

tdrwenski: It's a good question, I am not completely sure this is the right fix. The Grib1 version does not have this exception. And this function is getting called in the `Grib2CollectionBuilder::makeGroups`, so if you have a grib collection that you tell to recreate the index then that operation will fail due to this exception. It could be that there is a better fix for that case though.

haileyajohnson: yeah I'm not sure what the reason for immutability is here other than just a general migration toward immutable objects...

tdrwenski: I will have one more look if I can see a different way to fix this

tdrwenski: I reverted this change and moved the `timeUnitConverter` to the `Grib2CollectionBuilder`, and deprecated the one in the `Grib2Tables`. I think this is a much better solution. The `Grib2Table` still has various `getForecast...` methods that now no longer do time unit conversions, but actually the callers of these methods handle the time units already.

grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Tables.java
```diff
@@ -38,6 +38,7 @@
  * @since 4/3/11
  */
 @Immutable
+// TODO version 6: remove implements TimeUnitConverter
```
tdrwenski, line 41: Changing this inheritance now would I think technically be a breaking change that effects ~10 other classes that inherit from this one. If we decide in version 6 that this class isn't part of the public API then it is easy to fix.

haileyajohnson (approved):

PR #1111, opened 2022-11-08
AbstractIOServiceProvider.reacquire() should use getRaf() in order to…

WeatherGod:
… use the correct RandomAccessFile subclass

## Description of Changes
* Fixes a bug with FileCache of remote NetcdfFiles, particularly S3 files
* Added a unit test to check the hit/miss counts of the FileCache of an S3 file.
* Updated an existing unit test for FileCache to verify that hits occur for local files, too.

## PR Checklist

- [ ] Add labels
       until ready for review


WeatherGod:
This is "ready for review", in a sense, but I need someone on the UCAR network to verify that my new S3NetcdfFileCache test works.

Addresses one (of many) issues discovered in https://github.com/Unidata/tds/issues/311#issuecomment-1302476033

WeatherGod:
Ugh, further testing on my system shows that under semi-stressful loads, there seem to be some random loading errors. Not exactly sure what it means at this point.

```
2022-11-08T21:57:23.490 +0000 [     70343][      63] ERROR - uk.ac.rdg.resc.edal.wms.WmsServlet - Problem with GET request
java.lang.IllegalStateException: DataBTree doesnt start with TREE
	at ucar.nc2.iosp.hdf5.DataBTree$Node.<init>(DataBTree.java:168) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.iosp.hdf5.DataBTree$DataChunkIterator.<init>(DataBTree.java:128) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.iosp.hdf5.DataBTree.getDataChunkIteratorFilter(DataBTree.java:67) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.iosp.hdf5.H5tiledLayoutBB.<init>(H5tiledLayoutBB.java:97) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:222) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:200) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.NetcdfFile.readData(NetcdfFile.java:2122) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.Variable.reallyRead(Variable.java:817) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.Variable._read(Variable.java:768) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.Variable.read(Variable.java:600) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.Variable.read(Variable.java:574) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.ncml.AggDatasetOuter.read(AggDatasetOuter.java:310) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.ncml.AggregationOuter.reallyRead(AggregationOuter.java:418) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.ncml.AggregationNew.reallyRead(AggregationNew.java:28) ~[cdm-core-5.5.3.jar:5.5.3]
```

If I don't include this patch in my build of thredds, I don't see this error.

haileyajohnson:
Hey @WeatherGod ! Thanks for putting this PR together! I think we really need to dig down and figure out what the underlying issue is here and what the best way to address it is, and that might take us some time. We'll keep you update on what we figure out 😄

WeatherGod:
A possible reason to consider for the error could simply be that I'm back-porting the patches to a branch I have containing a bunch of my works that is off of the 5.5.3 tag. I then `publishToMavenLocal`, and copy the cdm-misc, cdm-s3, and cdm-core jar files over to place in the `webapps/thredds/WEB-INF/lib/` directory of the 5.4 docker image, and run that.

I know, it is a bit convoluted, but it works with the system I'm using. Could it be possible that doing something like this is subtly wrong and may possibly cause problems, perhaps?

WeatherGod:
Another possible clue. I do not get the above error if the S3 objects to be fetched are small-ish (or at least, I haven't been able to force an exception to occur yet for such datasets). I can very easily (but not always) force the exception when doing GetMap WMS calls for a larger dataset where each daily file is around 21MB.

WeatherGod:
If there is anything further I can do to help figure this out, please let me know. Should I keep investigating the cause of the exception I noted above (I have a couple of guesses), or are you all thinking of taking a completely different tack?

haileyajohnson:
@WeatherGod I'm still not sure what the answer is, and we haven't had time to dig in yet. I don't know whether this will be the approach we take, so I'm not sure if it's worth your time to keep going down this track.

WeatherGod:
Resuming my investigation of the problem revealed by this patch:
* So far, in every instance, exceptions happen upon `readData()` for objects greater than 256kb (the default remote buffer size) when reacquired from the file cache. The first load seems to always work, and never for smaller objects. Sometimes multiple reads succeed, but I believe (but not confirmed) that that is a case of multiple simultaneous first fetches of the same object due to a race condition in the file cache.
* I disabled the remote read cache of the RemoteRandomAccessFile, and the problem still happens.
* Most exceptions are of the variety I noted above, but a few times, I get a different one:
```
java.lang.NullPointerException: null
	at ucar.unidata.io.s3.S3RandomAccessFile.readRemote(S3RandomAccessFile.java:108) ~[cdm-s3-5.5.3.jar:5.5.3]
	at ucar.unidata.io.RemoteRandomAccessFile.read_(RemoteRandomAccessFile.java:102) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.unidata.io.RandomAccessFile.readBuffer(RandomAccessFile.java:512) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.unidata.io.RandomAccessFile.seek(RandomAccessFile.java:500) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.iosp.hdf5.DataBTree$Node.<init>(DataBTree.java:163) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.iosp.hdf5.DataBTree$DataChunkIterator.<init>(DataBTree.java:128) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.iosp.hdf5.DataBTree.getDataChunkIteratorFilter(DataBTree.java:67) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.iosp.hdf5.H5tiledLayoutBB.<init>(H5tiledLayoutBB.java:97) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:222) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:200) ~[cdm-core-5.5.3.jar:5.5.3]
	at ucar.nc2.NetcdfFile.readData(NetcdfFile.java:2125) ~[cdm-core-5.5.3.jar:5.5.3]
```

WeatherGod:
Also, I logged what 4-character value the H5 reader is getting instead of "TREE", and I'm probably not capturing it correctly: `x^�`  and `x^��`, whatever these mean.

haileyajohnson:
@WeatherGod I would suggest merging this PR into your own fork and testing if it fixes your issues

WeatherGod:
That's what I've been doing, hence my various reporting on the results of doing so. Yes, this fixes a significant caching issue, but it exposes a pre-existing, unrealized, bug in reacquistions of S3 objects. When I have some time, I'm going to try to make an SSCCE and open a new bug report about it. I'm curious if it impacts Http files, too. If it doesn't, then that would help narrow down the source of this bug.

But, as it stands, because this patch exposes a bug that throws exceptions, I can't use this patch in production yet.

tdrwenski:
We just [fixed an issue](https://github.com/Unidata/netcdf-java/pull/1122) that was [leading to errors](https://github.com/Unidata/tds/issues/293) like `DataBTree doesnt start with TREE`. Just noticed you are seeing those errors too. It was a race condition, so it could be that you were also encountering that and it was unrelated to your patch, just timing sensitive. Hopefully, this fix will resolve those errors for you as well.

WeatherGod:
Hmmm, I'll rebase and give it a try.

WeatherGod:
This is looking promising! I haven't hit those exceptions yet. The number of HttpRequests is relatively small compared to before, and memory usage appears to stay under control, perhaps (haven't really stressed this yet). There are still aspects of the cache that aren't fully optimal yet, but I think this is a good sign!

cdm/core/src/test/java/ucar/nc2/util/cache/TestNetcdfFileCache.java
```diff
@@ -68,8 +68,9 @@ private void loadFilesIntoCache(File dir, FileCache cache) {
   @Test
   public void testNetcdfFileCache() throws IOException {
     loadFilesIntoCache(new File(TestDir.cdmLocalTestDataDir), cache);
-
     cache.showCache(new Formatter(System.out));
+    assertThat(cache.hits.get()).isEqualTo(0);
+    assertThat(cache.miss.get()).isEqualTo(count);
```
WeatherGod, line 73: Did I correctly notice that you are all moving away from using `assertThat` now? Should I update my tests?

tdrwenski: Actually we are moving to only using the google truth asserts, so this is good as it.

WeatherGod:
I can't seem to figure out what the test failure is. It passes for me locally.

WeatherGod:
Oh, nevermind, found it. Odd that it isn't failing for me locally...

WeatherGod:
My current thinking is that one of the tests I expanded is actually timing-sensitive and might have behaved a bit differently between the two systems.

cdm/core/src/test/java/ucar/nc2/util/cache/TestNetcdfFileCache.java
```diff
@@ -148,6 +159,36 @@ public void testNetcdfFileCache() throws IOException {
...
+      }
+    }
+    for (FileCacheable ncfile : files) {
+      ncfile.close();
+    }
+
+    loadFilesIntoCache(new File(TestDir.cdmLocalTestDataDir), cache);
+    cache.showCache(new Formatter(System.out));
+    assertThat(cache.hits.get()).isEqualTo(saveCount);
+    assertThat(cache.miss.get()).isEqualTo(6 * saveCount);
```
WeatherGod: Alright, so what I think is happening here is that there is no guarantee at this point that all of the files are getting marked as unlocked before we try to reload the files into the cache, so then the hit count isn't what I'd expect. 

Any suggestions like a possible cache method call or using a short timer delay?

WeatherGod: I'm giving it a shot with doing a synchronous setting `file.isLocked` to false in the above for-loop. Also seems like I might be able to trigger this locally if I wait a long period of time between executions of the test. Might be a local disk caching introducing race conditions... yay!

WeatherGod:
I hope that kludge I added to the test works and that the current round of "passing" wasn't a fluke.

Also, a reminder that the test I added for testing the caching of S3 files is annotated as needing the UCAR network, like most of the other S3-related tests. Therefore, I have no way of confirming that it actually works. Someone on the inside will need to run it.

As a side note, has anyone considered utilizing Moto (mocked boto) to run a mini S3 service on localhost for some of these tests? http://docs.getmoto.org/en/latest/docs/server_mode.html I'm not familiar enough with java and gradle to know how to use it to spin up and spin down a moto_server process in the background of certain tests.

cdm/core/src/test/java/ucar/nc2/util/cache/TestS3NetcdfFileCache.java
```diff
@@ -0,0 +1,147 @@
...
+    // close all
+    List<FileCacheable> files = new ArrayList<>();
+    for (Object key : map.keySet()) {
+      FileCache.CacheElement elem = map.get(key);
+      assertThat(elem.list.size()).isEqualTo(1);
+      for (FileCache.CacheElement.CacheFile file : elem.list) {
+        // Need to do it this way instead of directly closing them
+        // because closing the files changes the iterations
+        files.add(file.ncfile);
+      }
```
WeatherGod: Oh, I probably should add the same synchronization trick I had in the other test here too.

WeatherGod:
just managed to make it fail locally with full logging. So that kludge wasn't sufficient.

WeatherGod:
I can't consistently reproduce the test failure. There are times when I can reproduce the failures one after another, and then there are times where I could run it 5 times in a row and get successes. When it fails, it looks like the cache was in the middle of pruning/scouring the cache, even though just a few seconds have passed. Other times, I purposely put in a 10 second wait after closing the files and everything would still be there after the sleep.

WeatherGod:
So, I modified these tests for my own purposes, thinking it would help ensure that the cache works as expected. However, they don't actually directly test the change here (at least, not the one that actually runs in CI/CD). At the moment, I'm inclined to remove the last part of the two cache tests and make a separate issue noting how the behavior of the cache isn't quite as expected.

Agreed?

WeatherGod:
Is it required that I do another merge of dev to this branch?

WeatherGod:
I have been using this patch in my own system for about a month and a half now, and I have not experienced any exceptions due to it, and performance is noticeably better, particularly with memory usage. I think this is should go in.

cdm/core/src/test/java/ucar/nc2/util/cache/TestS3NetcdfFileCache.java
```diff
@@ -0,0 +1,122 @@
...
+import ucar.nc2.dataset.NetcdfDatasets;
+import ucar.nc2.util.CancelTask;
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.util.*;
+import ucar.unidata.util.test.category.NeedsUcarNetwork;
+import ucar.unidata.util.test.category.NeedsExternalResource;
+
+/** Test FileCache. */
+@Category(NeedsExternalResource.class)
```
tdrwenski: This may be used in some of our tests but should not be necessary in order to perform S3 tests. We have a `thredds-test-data` public test bucket that should work for this. For instance, if you just need a single s3 file you could use the key `"cdms3:thredds-test-data?testData.nc"`

cdm/core/src/test/java/ucar/nc2/util/cache/TestS3NetcdfFileCache.java
```diff
@@ -0,0 +1,122 @@
+/*
+ * Copyright (c) 1998-2022 University Corporation for Atmospheric Research/Unidata
+ * See LICENSE for license information.
+ */
+package ucar.nc2.util.cache;
```
tdrwenski: I don't think you can run the S3 tests in this package because in this package you won't be able to open S3 files. I think it will have to be moved to `package ucar.unidata.io.s3;`. This is because our S3 dependencies are not in the cdm core package and the S3RandomAccessFile.Provider only gets dynamically loaded in the S3 package.

Moving these test to another package may require rewriting them a bit because I see some package private functions being used here (like `cache.hits`). If you just want to see what is in the cache that is public (`RandomAccessFile.getGlobalFileCache().showCache()`) so maybe these tests could be simplified. If it is really necessary to check the hits/misses to test this issue we could consider adding a public method to the cache for that.

tdrwenski:
@WeatherGod It is great to hear that you have been testing with this fix and it has helped performance! I had a few remarks on improving the tests before we merge this. If you would like us to work on the tests instead, let us know!

cdm/core/src/test/java/ucar/nc2/util/cache/TestS3NetcdfFileCache.java
WeatherGod: ah, I'll use that, thanks!

cdm/core/src/test/java/ucar/nc2/util/cache/TestS3NetcdfFileCache.java
WeatherGod: Since the original bug was that the requests were silently missing the cache, I think it is important to have a test that verifies that the cache is performing as expected. This means we need some sort of way to verify the cache state, whether it be by counting the hits and misses, or having a list of the cache keys to test against. I'll leave it up to you devs to decide what to expose as read-only public. I can adjust accordingly.

Good point about being in the wrong location. I'm not used to developing Java libraries, so I am not used to thinking about such considerations.

tdrwenski: I think actually the existing functions should be usable for your tests (if a bit clunky), so probably no need to add extra functionality there. To see a list of entries in the cache you can do `RandomAccessFile.getGlobalFileCache().showCache()`. To check the stats (as a string like "hits= 0 miss= 0 nfiles= 0 elems= 0") you can something like
```
    final FileCacheIF rafFileCache = RandomAccessFile.getGlobalFileCache();
    final Formatter formatter = new Formatter();
    rafFileCache.showStats(formatter);
    assertThat(formatter.toString()).contains("miss= " + size);
```

WeatherGod: Great suggestion. I went ahead and redesigned it. Still think it would be a good idea to eventually consider utilizing something like `moto` to  provide a mocked S3 service, but that is obviously outside the scope of this PR.

tdrwenski: Thanks for the suggestion, I agree that would be nice!

tdrwenski (approved):
Looks good, thank you!

Issue #199, opened 2020-02-03
Presence of libjnidispatch.jnilib triggers Mac notarization failure


rschmunk:
I am trying to finish packaging up a copy of Panoply for macOS using netCDF-Java 5.3 SNAPSHOT. Everything is great right up until almost the end when I have to get the disk image DMG notarized by Apple. That fails because of the presence of libjnidispatch.jnilib in the NJ jar, i.e., netcdfAll-5.3.0-SNAPSHOT.jar/com/sun/jna/darwin/libjnidispatch.jnilib. In short, the notarization process is complaining that the libjnidispatch.jnilib binary is not properly signed.

Note that this problem did not occur last Friday (January 31) when I was making a Mac package, but I have a dim recollection that Apple was going to be closing some loophole for notarizing Java based apps effective February 1. Or perhaps that was apps built on an an old Java, such as the Java 8 that NJ is based on.

I realize that dealing with this is pretty much outside Unidata's purview, but it's something that could completely break my ability to distribute a netCDF-Java based app to macOS users with an up-to-date operating system. But I am wondering what libjnidispatch.jnilib is used for in NJ, and whether it's really necessary?

lesserwhirls:
Greetings @rschmunk! Yes, the `libjnidispatch.jnilib` is coming in from the `netcdf4` module, which writes netCDF-4 files via the C library, and can also read netCDF files through the C library for debugging purposes. It's actually JNA that is pulling in `libjnidispatch.jnilib`. That code should never be touched unless the netCDF-Java is explicitly told that netCDF-C exists and should be used, so it should be safe to remove.

It looks like Apple [flipped the switch](https://developer.apple.com/news/?id=12232019a) on their relaxed notarization prerequisites, and that's causing lots of headaches. I'm not sure if you ran across this [stackoverflow post](https://stackoverflow.com/a/53528020), but it describes what they had to do to sign native libraries bundled within a jar.

rschmunk:
@lesserwhirls, So if `.unidata/nj22Config.xml` has been read in so that the user can use `libnetcdf`, would that jnilib be needed? As some point a few years ago due to a user request, I modified Panoply to look for that XML file and if found, then feed it to RuntimeConfigParser.

lesserwhirls:
In that case yes, `libjnidispatch.jnilib` would be needed as it allows the JNA library to make calls to the netCDF-C library.

rschmunk:
[...]

PR #1116, opened 2022-11-28
Update JNA version

ethanrd:
## Description of Changes

For JNA before version 5.7.0, use of JNA fails at runtime on MacOS with M1 chip.

This PR updates to current JNA version 5.12.1.

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/netcdf-java-platform/build.gradle
+++ b/netcdf-java-platform/build.gradle
@@ -28,7 +28,7 @@ dependencies {
     // netcdf4, dap4:d4lib
     // Tricky dependency here. We need to make sure that we keep in-line with the version
     // that chronicle-map uses in the TDS, or else we see bad things happen on the TDS side.
-    api 'net.java.dev.jna:jna:5.4.0'
+    api 'net.java.dev.jna:jna:5.12.1'
 
     // Annotations: Nullable
     api 'com.google.code.findbugs:jsr305:3.0.2'
--- a/netcdf-java-platform/build.gradle
+++ b/netcdf-java-platform/build.gradle
@@ -28,7 +28,7 @@ dependencies {
     // netcdf4, dap4:d4lib
     // Tricky dependency here. We need to make sure that we keep in-line with the version
     // that chronicle-map uses in the TDS, or else we see bad things happen on the TDS side.
-    api 'net.java.dev.jna:jna:5.8.0'
+    api 'net.java.dev.jna:jna:5.12.1'
 
     // Annotations: Nullable
     api 'com.google.code.findbugs:jsr305:3.0.2'
--- a/netcdf-java-platform/build.gradle
+++ b/netcdf-java-platform/build.gradle
@@ -28,7 +28,7 @@ dependencies {
     // netcdf4, dap4:d4lib
     // Tricky dependency here. We need to make sure that we keep in-line with the version
     // that chronicle-map uses in the TDS, or else we see bad things happen on the TDS side.
-    api 'net.java.dev.jna:jna:5.12.1'
+    api 'net.java.dev.jna:jna:5.8.0'
 
     // Annotations: Nullable
     api 'com.google.code.findbugs:jsr305:3.0.2'


lesserwhirls:
How will this play out with the version used by Chronicle map on the TDS-side?

lesserwhirls:
Here is an example describing where things went sideways when there was a version mismatch: https://github.com/Unidata/tds/pull/37

lesserwhirls:
In happier news though, the update would address https://github.com/Unidata/netcdf-java/issues/199 as JNA 5.7.0 removed the native library code for darwin x86.

ethanrd:
Hi @lesserwhirls - I ran into this failure while building/testing TDS on a new M1 machine. The symptoms where similar to what you described in Unidata/tds#37: `500` on any service except `catalog` and there were messages about null `datasetMap` like you mentioned and several related errors including a `NoClassDefFoundError` for JNA:

```
2022-11-16T15:29:11.957-0700 [     70500][        ] ERROR serverStartup: TdsInit: Failed initialize GridDatasetInv cache= /Users/edavis/code/tds/tds/src/test/content/thredds/cache/collection/
net.openhft.chronicle.hash.ChronicleHashRecoveryFailedException: java.lang.NoClassDefFoundError: Could not initialize class com.sun.jna.Native
	at net.openhft.chronicle.map.ChronicleMapBuilder.openWithExistingFile(ChronicleMapBuilder.java:1926) ~[chronicle-map-3.21ea6.jar:3.21ea6]
	at net.openhft.chronicle.map.ChronicleMapBuilder.createWithFile(ChronicleMapBuilder.java:1716) ~[chronicle-map-3.21ea6.jar:3.21ea6]
        ...
```

I found an [issue](https://github.com/oshi/oshi/issues/1462) (from Jan 2021) that seemed relevant, "Apple M1 hardware incompatibility for JNA Native class", and [mentions](https://github.com/oshi/oshi/issues/1462#issuecomment-768676937) the problem being fixed in JNA 5.7.0.

Since bumping the JNA version to "5.12.1" in `netcdf-java-platform/build.gradle`, my local build of TDS 5 using the netCDF-Java with the updated JNA no longer has failures on non-catalog requests or the various error messages.

Also, it looks like TDS now inherits the JNA version from the declaration in `netcdf-java-platform/build.gradle` rather than in `tds/gradle/any/dependencies.gradle` as in Unidata/tds#37. So maybe the potential conflict is gone now that the JNA version is only declared in one place?

tdrwenski:
If it is helpful, I can do some testing on the chronicle map cache with this JNA version (was working on testing that recently anyhow).

lesserwhirls:
The issue isn't so much the version of JNA that the TDS uses, it's the version that `chronical-map` uses. The version of `chronical-map` that the TDS is pulling in (`3.21ea6` from `chronical-bom` `2.21ea74`) currently uses JNA `5.5.0`. So the question is, how well will chronical-map handle having JNA version `5.5.0` swapped out with `5.12.1` at runtime?

We don't use much of the features of `chronical-map`, so my guess is it will fail spectacularly as before (tons of 500s), but it will need to be tested on multiple platforms to be sure. I'd be happy to give it a spin on Windows 10 if no devs are running Windows.

Longer term, we should probably look for a replacement for `chronical-map`. It was used to help with ESGF performance issues, but since they are moving away from the TDS, maybe it does not matter at this point given the headaches we've had.

ethanrd:
Ah, yes, I see. And, unfortunately, the current chronicle-map still uses JNA 5.5.0. Except on mac os, if I'm reading [this](https://repo1.maven.org/maven2/net/openhft/third-party-bom/3.22.3/third-party-bom-3.22.3.pom) right, where it uses 5.8.0.

lesserwhirls:
> Ah, yes, I see. And, unfortunately, the current chronicle-map still uses JNA 5.5.0. Except on mac os, if I'm reading [this](https://repo1.maven.org/maven2/net/openhft/third-party-bom/3.22.3/third-party-bom-3.22.3.pom) right, where it uses 5.8.0.

That's correct, but that is selected at build time, so only active if building on a mac. So in order to keep JNA versions between the TSD and netCDF-Java in-line (which we may or may not need, depending on testing) while also supporting M1 chips, we'd need to 1) update netCDF-Java to use JNA version `5.8.0` and build/publish a snapshot, 2) update the TDS to use the latest `chronicle-bom` (version `2.24ea9`) and latest snapshot of netCDF-Java, and 3) build the TDS war on a Mac -or- have Gradle ensure the TDS selects JNA `5.8.0` when resolving dependencies.

ethanrd:
Sounds good. I'll start testing and will update this PR to use JNA 5.8.0 instead of 5.12.1.

dbwiddis:
Hi, a JNA maintainer here.
> So the question is, how well will chronical-map handle having JNA version 5.5.0 swapped out with 5.12.1 at runtime?

JNA follows [semver](https://semver.org/), so I don't think this should cause any problems staying within 5.x.  I just reviewed the [change log](https://github.com/java-native-access/jna/blob/master/CHANGES.md) and don't see anything that would be risky at all, and several bug fixes/improvements between those versions.   The previous version mismatch issues linked earlier in this thread look like a 4.x to 5.x change.

> In happier news though, the update would address https://github.com/Unidata/netcdf-java/issues/199 

Yep, you definitely need 5.6.0+ for macOS 11+ compatibility.  If chronical-map is still on 5.5.0 it is incompatible with macOS 11, 12, and 13.

lesserwhirls:
Thank you for reaching out @dbwiddis! Sounds like going with 5.12.1 will be ok @ethanrd.

ethanrd:
Sounds good! I had dropped back to 5.8.0 just to match what Chronicle currently uses. But I will bump back to 5.12.1 and finish up some testing.

Thanks! @dbwiddis and @lesserwhirls

haileyajohnson (approved):

PR #1128, opened 2023-01-04
Fix VariableDS issue in mergeNcml

tdrwenski:
## Description of Changes

With a datasetScan with ncml, errors due to `Variable`s in a `NetcdfDataset` not being instances of `VariableDS` occurred. My fix for this is to use a different constructor for `NetcdfDataset` in `mergeNcml`. In addition, I deprecated the constructor previously used there, it basically just constructs the parent class `NetcdfFile` and isn't used anywhere else.

In addition to testing `mergeNcml`, I added a few tests for the other constructors of `NetcdfDataset`. I also added a couple of tests in the [TDS](https://github.com/Unidata/tds/pull/335).

A lingering additional question I have is if we should fix the `NetcdfDataset` builder to not allow `Variable`s that are not `VariableDS` (see TODO in TestNetcdfDatasetBuilder.java). This situation can cause similar errors when a dataset is enhanced but I am not sure if/how we should fix this.

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/core/src/test/java/ucar/nc2/internal/ncml/TestNcmlReader.java
+++ b/cdm/core/src/test/java/ucar/nc2/internal/ncml/TestNcmlReader.java
@@ -20,7 +20,7 @@ import ucar.unidata.util.test.TestDir;
 public class TestNcmlReader {
 
   @Test
-  public void shouldMergeNcml() throws IOException {
+  public void shouldMergeNcml() throws IOException, JDOMException {
     final String filename = TestDir.cdmLocalTestDataDir + "example1.nc";
 
     try (NetcdfFile netcdfFile = NetcdfDatasets.openFile(filename, null)) {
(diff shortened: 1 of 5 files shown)


cdm/core/src/test/java/ucar/nc2/internal/ncml/TestNcmlReader.java
```diff
@@ -0,0 +1,51 @@
...
+  private static Element getNcmlElement() {
+    final String ncml = TestNcmlRead.topDir + "modifyVars.xml";
+
+    try {
+      SAXBuilder saxBuilder = new SAXBuilder();
+      saxBuilder.setExpandEntities(false);
+      Document jdomDoc = saxBuilder.build(ncml);
+      return jdomDoc.getRootElement();
+    } catch (JDOMException | IOException e) {
+      Assert.fail();
```
haileyajohnson: Pedantic but I'd love to not use this `Assert.fail()` here

haileyajohnson (approved):

cdm/core/src/test/java/ucar/nc2/internal/ncml/TestNcmlReader.java
tdrwenski: Good point, that probably shouldn't be used here. I appreciate pedantic! :)

tdrwenski: Resolved by not catching that exception in the test

PR #1137, opened 2023-01-19
Port validate for CatalogBuilder

oxelson:
Added port validation in CatalogBuilder:
- root-privileged ports under 1024 not allowed unless 80 or 443

Added unit tests for aforementioned port validation

### DIFF
--- a/cdm/core/src/main/java/thredds/client/catalog/builder/CatalogBuilder.java
+++ b/cdm/core/src/main/java/thredds/client/catalog/builder/CatalogBuilder.java
@@ -59,12 +59,28 @@ public class CatalogBuilder {
     }
 
     this.baseURI = baseURI;
+    try {
+      validatePort(baseURI);
+    } catch (IllegalArgumentException e) {
+      errlog.format("Invalid port number: ", baseURI.toASCIIString(), e.getMessage());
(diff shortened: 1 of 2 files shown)


cdm/core/src/main/java/thredds/client/catalog/builder/CatalogBuilder.java
```diff
@@ -59,12 +59,28 @@ public Catalog buildFromLocation(String location, URI baseURI) {
     }
 
     this.baseURI = baseURI;
+    try {
+      validatePort(baseURI);
+    } catch (IllegalArgumentException e) {
+      errlog.format("Invalid port number = '%s' err='%s'%n ", baseURI.toASCIIString(), e.getMessage());
```
haileyajohnson, line 65: I think this will end up returning port information in the browser error message...

oxelson: It should be OK, as it's not giving away any sensitive info. It's send this if they try a root-privileged port that isn't 80 or 443... which is probably deliberate attempt at shenanigans.

haileyajohnson: works for me

haileyajohnson (approved):

PR #1144, opened 2023-02-03
Modify and Delete attributes for aggregations with scans

### LINKED ISSUE #1065: [5.x]: Ncml remove attribute fails for aggregation with scan

v5.x

### What went wrong?

Removing and renaming attributes in ncml fails for an aggregation with a scan element. Found in https://github.com/Unidata/tds/issues/268

For example, in the snippet below `long_name` is not removed. In addition, uncommenting the `renamed_units` line gives an error that the ncml cannot be read.
```
<?xml version="1.0" encoding="UTF-8"?>
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2">

  <variable name="T">
    <attribute name="new_attribute" type="string" value="new attribute value"/>
<!--    <attribute name="renamed_units" orgName="units"/> TODO this causes an error when reading the ncml-->
    <remove type="attribute" name="long_name"/>
  </variable>

  <aggregation dimName="time" type="joinExisting">
    <scan location="nc/" regExp="^(jan.nc|feb.nc)$"/>
  </aggregation>

</netcdf>
```

However, if we replace the `aggregation` with
```
  <aggregation dimName="time" type="joinExisting">
    <netcdf location="nc/jan.nc"/>
    <netcdf location="nc/feb.nc"/>
  </aggregation>
```
both the removal and the renaming of attributes works as expected.

### Relevant stack trace

_No response_

### Relevant log messages

_No response_

N/A

### Code of Conduct

mnlerman:
## Description of Changes

resolves #1065 

Aggregations with scans didn't handle removing or renaming attributes. I added code to handle removing attributes and added an extra condition when reading the attribute to check to see if already existed in the attribute container. Even though there isn't a reference variable, the rest of the function executes correctly by reading the attribute element and renaming the existing attribute in the container 

## PR Checklist
- [x] Add labels
       until ready for review


cdm/core/src/main/java/ucar/nc2/internal/ncml/NcmlReader.java
```diff
@@ -665,13 +665,22 @@ private void readAtt(AttributeContainer dest, @Nullable AttributeContainer ref,
     boolean newName = (nameInFile != null) && !nameInFile.equals(name);
     if (nameInFile == null) {
       nameInFile = name;
-    } else if (null == findAttribute(ref, nameInFile)) { // has to exists
-      errlog.format("NcML attribute orgName '%s' doesnt exist. att=%s in=%s%n", nameInFile, name, refName);
-      return;
+    } else if (findAttribute(ref, nameInFile) == null) {
+      if (findAttribute(dest, nameInFile) == null) { // has to exist
+        errlog.format("NcML attribute orgName '%s' doesnt exist. att=%s in=%s%n", nameInFile, name, refName);
```
haileyajohnson: super nit but I'd make these one condition with an &&, since there's nothing else inside the `else if`

cdm/core/src/test/java/ucar/nc2/internal/ncml/TestNcmlModifyAtts.java
```diff
@@ -258,4 +268,70 @@ public void testReadData2() throws IOException {
     Assert2.assertNearlyEquals(dataI.getDoubleNext(), 4.0);
     Assert2.assertNearlyEquals(dataI.getDoubleNext(), 2.0);
   }
+
+  @Test
```
haileyajohnson, line 272: Would you mind refactoring these to have a helper method that takes the filename and does all the verification to avoid duplication? :) Otherwise looks great

haileyajohnson (changes requested):

cdm/core/src/test/data/ncml/aggregationScanModifyAttributes.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2">
+
+  <variable name="T">
+    <attribute name="new_attribute" type="string" value="new attribute value"/>
+    <attribute name="renamed_units" orgName="units"/> <!--TODO this causes an error when reading the ncml-->
```
tdrwenski: This TODO can now go
```suggestion
    <attribute name="renamed_units" orgName="units"/>
```

mnlerman: Ahh yep, I missed that! Removed. Thanks :)

haileyajohnson (approved):

cdm/core/src/test/java/ucar/nc2/internal/ncml/TestNcmlModifyAtts.java
mnlerman: Yeah, I just refactored and pushed it again. Thanks for the input! I didn't catch that they ended up being the same. Let me know what you think :)

PR #1151, opened 2023-02-26
Fix EnumTypedef name problem - Attempt 2

DennisHeimbigner:
## Description of Changes
re: Issue https://github.com/Unidata/netcdf-java/issues/126

The code for handling enum types in Group.java is
incorrect. When creating a new enum type, it is appropriate to
search only the current group for a conflicting name and this is
what the current code does.  But when an enum typed variable
searches for the matching enum type, it must search not only the
current group but all parent groups and the current code does
not do that.

The fix consists of two similar parts.
1. Modify Group.findEnumeration to take an extra boolean parameter
to control if search is this group only or to search up the group's parents.
2. Modify Group.Builder.findEnumTypedef to act like part 1 but to search
the sequence of parent Group.Builder instances.

As a consequence, this PR modifies a number of other files to
adapt to the modified signatures.

## Note
This same problem appears to affect the opaque type also, but
fixing that may be a bit more difficult because CDM appears to
convert opaque types to byte typess.

## PR Checklist

- [x] Add labels
       until ready for review

## Addendum 1
It turns out that I missed the error in the code in H5headerNew that
attempts to match an enum typed variable with the proper enumeration
type declaration.

The problem and fix (as described by a comment in H5headerNew)
is a bit of a hack. It should be fixed in H5Object.read().
Unfortunately, information is not being passed down so that
the proper fix can be applied early in the HDF5->CDM translation.
Fixing this would affect a lot of function signatures.

Also modified TestEnumTypedef.java to test two cases:
1. the actual enum type def is in the same group as the variable that uses it.
2. the actual enum type def is in a parent group of the variable that uses it

## Misc. Other Changes
* Suppress printing of _NCProperties to simplify text-based comparison testing.

## Addendum 2 (12/19/2022)
* NetcdfFile.java: convert to use H5iospNew (needed by TestH5iosp.java)
* H5headerNew.java: provide get function for accessing the btree (needed by TestDataBTree.java)
* H5iospNew.java: make getRandomAccessFile() method public (needed by tests)
* CompareNetcdf2.java: Add a constructor to specify if attribute name comparison should ignore case or not. It turns out that some tests require case sensitive name matching. Specifically TestCoordSysCompare.java and TestN3iospCompare.java

## Addendum 3 (2/25/2023)
Make additional changes to fix Jenkins failures.

* Revert Group.attributes()
* Revert the use of H5iospNew in NetcdfFile.
* Add to Group a method that looks for a matching enumeration purely structurally, as opposed to name matching. This corresponds to the same method in Group Builder.
* Rebuild the EnumTypedef search algorithm in H5headerNew and H5header to handle cases in Jenins only enum tests.
* Revert various calls to Group.findEnumeration().
* Add some test case data: ref_anon_enum.h5 and test_enum_type.nc
* Convert TestEnumTypedef to JUNIT.parameter form.
* Revert TestH5iosp.
* The URL used in TestSequence is no longer valid, so change to an equivalent url on the remotetest server.


cdm/core/src/main/java/ucar/nc2/Group.java
```diff
@@ -374,20 +376,56 @@ public boolean removeAttributeIgnoreCase(String attName) {
...
+    }
+    if (ed != null)
+      return ed;
+    // Optionally search parents
+    if (searchup) {
+      Group gb = getParentGroup();
+      if (gb != null)
+        ed = gb.findSimilarEnumTypedef(template, searchup);
+      if (ed != null)
+        return ed;
```
haileyajohnson: Nitpicky but I don't think this conditional is necessary since the function defaults to returning null anyway

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
```diff
@@ -1601,7 +1649,7 @@ public class Vinfo {
 
     // chunked stuff
     boolean isChunked;
-    DataBTree btree; // only if isChunked
+    public DataBTree btree; // only if isChunked
```
haileyajohnson: this doesn't need to be public if we're adding a getter and don't want it to be publicly settable

haileyajohnson (approved):

cdm/core/src/main/java/ucar/nc2/iosp/hdf5/H5header.java
```diff
@@ -1739,18 +1739,83 @@ private boolean makeVariableShapeAndType(Variable v, MessageDatatype mdt, Messag
...
+
+      if (actualEnumTypedef == null) {
+        log.warn("Missing EnumTypedef: {}", mdt.enumTypeName);
+        throw new IllegalStateException("Missing EnumTypedef: " + mdt.enumTypeName);
+      }
+
+      // associate with the variable
+      v.setEnumTypedef(actualEnumTypedef);
+      /*
+       * Group ncGroup = v.getParentGroupOrRoot();
```
tdrwenski: maybe the commented out block can go now

cdm/core/src/main/java/ucar/nc2/Group.java
```diff
@@ -1114,6 +1152,62 @@ public Group.Builder commonParent(Group.Builder other) {
...
+    /**
+     * Find a Enumeration in this or a parent Group Builder, using its short name.
+     * 
+     * @param name for which to search
+     * @param searchup if true, then search this group and then parent groups.
+     */
+    public Optional<EnumTypedef> findEnumTypedef(String name, boolean searchup) {
+      if (name == null)
+        return Optional.empty();
+      // name = NetcdfFile.makeNameUnescaped(name);
```
tdrwenski: maybe this line can go now

cdm/core/src/main/java/ucar/nc2/Group.java
```diff
@@ -374,20 +376,56 @@ public boolean removeAttributeIgnoreCase(String attName) {
...
+   * @param searchup if true, then search this group and then parent groups.
+   */
+  public EnumTypedef findSimilarEnumTypedef(EnumTypedef template, boolean searchup) {
+    EnumTypedef ed = null;
+    assert (template != null);
+    // search this group builders's EnumTypedefs but with constraint on name
+    {
+      Optional<EnumTypedef> edopt = this.enumTypedefs.stream().filter(e -> (e.equalsMapOnly(template))).findFirst();
+      ed = (edopt.isPresent() ? edopt.get() : null);
+    }
```
tdrwenski: I think this can be simplified to
```suggestion
    assert (template != null);
    // search this group builders's EnumTypedefs but with constraint on name
    EnumTypedef ed = enumTypedefs.stream().filter(e -> e.equalsMapOnly(template)).findFirst().orElse(null);
```

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
```diff
@@ -1553,19 +1547,73 @@ private boolean makeVariableShapeAndType(Group.Builder parent, Variable.Builder
...
+
+      EnumTypedef actualEnumTypedef = null; // The final chosen EnumTypedef
+
+      // Step 1:
+      // See if an independent enum type already exists with the same name
+      Optional<EnumTypedef> candidate = parent.findEnumTypedef(mdt.enumTypeName, true);
+      if (candidate.isPresent()) {
+        // There is an independent type, so use it.
+        actualEnumTypedef = candidate.get();
+      }
```
tdrwenski: This this and the one below on 1584 could be simplified to:
```suggestion
      actualEnumTypedef = candidate.orElse(null);
```

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
```diff
@@ -1553,19 +1547,73 @@ private boolean makeVariableShapeAndType(Group.Builder parent, Variable.Builder
...
+          newname = mdt.enumTypeName + "_enum_t";
+        } else {
+          newname = mdt.enumTypeName;
+        }
+        actualEnumTypedef = new EnumTypedef(newname, mdt.map);
+        // Add to the current group(builder)
+        parent.addEnumTypedef(actualEnumTypedef);
+      }
+
+      if (actualEnumTypedef == null) {
```
tdrwenski: I don't think this `if` can ever be true since above if `actualEnumTypedef == null` then it gets constructed

cdm/core/src/main/java/ucar/nc2/Group.java
DennisHeimbigner: Done.

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
DennisHeimbigner: Done.

cdm/core/src/main/java/ucar/nc2/iosp/hdf5/H5header.java
DennisHeimbigner: Done

cdm/core/src/main/java/ucar/nc2/Group.java
DennisHeimbigner: Done

cdm/core/src/main/java/ucar/nc2/Group.java
DennisHeimbigner: done

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
DennisHeimbigner: done

cdm/core/src/main/java/ucar/nc2/internal/iosp/hdf5/H5headerNew.java
DennisHeimbigner: done

tdrwenski (approved):

tdrwenski:
Thanks for making those changes Dennis! Is this ready to merge now?

DennisHeimbigner:
Actually, let me run it with Jenkins one last time.

DennisHeimbigner:
ok, it still passes Jenkins, so it is ready to merge.

PR #1154, opened 2023-03-08
Add method to combine Arrays

tdrwenski:
## Description of Changes

Add `Array.factoryCopy` method that combines multiple Arrays. This functionality is needed for the gCDM, where the data may need to be chunked and then recombined. As I made some simple tests for this function, I thought it might be nice to make a separate PR for it.

Also added tests for some of the `Array.factory` methods (not for all of them though). Also added a TODO to the `DataType.Object` which seems like a hack that we could revisit.

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/core/src/main/java/ucar/ma2/Array.java
+++ b/cdm/core/src/main/java/ucar/ma2/Array.java
@@ -388,7 +388,8 @@ public abstract class Array {
   public static Array factoryCopy(DataType dataType, int[] shape, List<Array> arrays) {
     if (arrays.isEmpty()) {
       throw new IllegalArgumentException("Expected a non-empty list of Arrays to combine");
-    } else if (arrays.size() == 1) {
+    }
+    if (arrays.size() == 1) {
       return factory(dataType, shape, arrays.get(0).getStorage());
     }
     final long size = Index.computeSize(shape);
--- a/cdm/core/src/main/java/ucar/ma2/Array.java
+++ b/cdm/core/src/main/java/ucar/ma2/Array.java
@@ -377,6 +377,55 @@ public abstract class Array {
     return factory(org.getDataType(), shape, org.getStorage());
   }
 
+  /**
+   * Combine list of Arrays by copying the underlying Arrays into a single primitive array
+   *
+   * @param dataType the DataType
+   * @param shape the shape of the combined array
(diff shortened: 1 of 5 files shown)


cdm/core/src/main/java/ucar/ma2/Array.java
```diff
@@ -377,6 +377,55 @@ public static Array makeArrayRankPlusOne(Array org) {
...
+   *
+   * @param dataType the DataType
+   * @param shape the shape of the combined array
+   * @param arrays non-empty list of arrays of dataType to combine
+   * @return a new Array containing data from the arrays
+   */
+  public static Array factoryCopy(DataType dataType, int[] shape, List<Array> arrays) {
+    if (arrays.isEmpty()) {
+      throw new IllegalArgumentException("Expected a non-empty list of Arrays to combine");
+    } else if (arrays.size() == 1) {
```
haileyajohnson: this doesn't need to be an `else if` :P

cdm/core/src/test/java/ucar/ma2/TestArray.java
```diff
@@ -0,0 +1,121 @@
...
+    final int[] combinedShape = new int[] {2};
+    final List<Array> arrays = new ArrayList<>();
+    arrays.add(arrayStructure);
+    arrays.add(arrayStructure);
+
+    final ArrayStructure combinedArray = (ArrayStructure) Array.factoryCopy(DataType.STRUCTURE, combinedShape, arrays);
+
+    assertThat(combinedArray.getShape()).isEqualTo(combinedShape);
+    assertThat(combinedArray.getArray(0, members.getMember(0))).isEqualTo(array);
+    assertThat(combinedArray.getArray(1, members.getMember(0))).isEqualTo(array);
```
haileyajohnson: Will this work if the `ArrayStructures` don't have the same members?

tdrwenski: Yeah good point, it probably would fail in some way, but now I added an explicit check with an exception for this situation. I also added a couple more tests for the `ArrayStructures`. In order to check the `StructureMembers` are the same I also added an `StructureMembers::equals` and tests for that.

haileyajohnson (approved):

PR #1155, opened 2023-03-16
Port gcdm from v6 to v5

tdrwenski:
## Description of Changes

gCDM (gRPC for CDM) port from v6 to v5:
- The gCDM server which can send netcdf files.
- The client, which is used for testing, is `GcdmNetcdfFile` which  `extends NetcdfFile`.
- The `GcdmConverter` which encodes (turns an object into a proto object) and decodes (other way around)
- Tests

Notes:
- structures with vlens don't work correctly in netcdf-java so are ignored here
- sequences are not well tested-- they only exist in some bufr test files and cannot be written to netcdf 3/4 so don't know how to make test files for that
- structures with large members are not chunked correctly (I just added a TODO for that)
- I changed the `CompareNetcdf2.java` to avoid OutOfMemory errors. These occurred even when I did not use the gCDM, when a variable is very large and the comparison code reads it twice in memory.

## PR Checklist

- [ ] Add labels
       until ready for review


cdm/gcdm/build.gradle
```diff
@@ -0,0 +1,88 @@
...
+  testImplementation project(':cdm-test-utils')
+  testImplementation 'com.google.truth:truth'
+  testImplementation 'junit:junit'
+  testImplementation 'commons-io:commons-io'
+  testImplementation 'org.mockito:mockito-core'
+  testImplementation "io.grpc:grpc-testing:${depVersion.grpc}"
+
+  // logging
+  implementation 'org.slf4j:slf4j-api'
+  runtimeOnly 'ch.qos.logback:logback-classic'
```
haileyajohnson, line 48: do we want logback in the runtime or just the test runtime?

cdm/gcdm/src/main/java/ucar/gcdm/GcdmConverter.java
```diff
@@ -0,0 +1,733 @@
...
+      final Data inner = vlenData.getVlenData(i);
+      storage[i] = decodeData(inner);
+    }
+
+    return Array.makeVlenArray(shape, storage);
+  }
+
+  ////////////////////////////////////////////////////////////////
+
+  public static GcdmNetcdfProto.DataType convertDataType(DataType dataType) {
```
haileyajohnson, line 616: I think this is fine for now, but there's gotta be a better way to do it... two hashmaps?

cdm/gcdm/src/test/java/ucar/gcdm/TestGcdmNetcdfFileProblem.java
haileyajohnson, line 1: I know you didn't name this, but I kinda hate this naming scheme, it's not very descriptive

cdm/gcdm/src/test/java/ucar/gcdm/TestGcdmNetcdfFileProblem.java
```diff
@@ -0,0 +1,371 @@
...
+      CompareNetcdf2 compareNetcdf2 = new CompareNetcdf2(formatter, false, true, true);
+      boolean ok = compareNetcdf2.compareVariable(ncVar, gcdmVar, CompareNetcdf2.IDENTITY_FILTER);
+      logger.debug(formatter.toString());
+      assertThat(ok).isTrue();
+    }
+  }
+
+  private static void writeNetcdfFile(String filename, String variableName, DataType dataType, Array values)
+      throws IOException, InvalidRangeException {
+    NetcdfFormatWriter.Builder writerBuilder = NetcdfFormatWriter.createNewNetcdf3(filename);
```
haileyajohnson: Can this be refactored so that instead of passing filename in, the tempfolder creates a file in this method? As an extra safety catch against making persistent files in tests.

cdm/gcdm/src/test/java/ucar/gcdm/TestGcdmNetcdfFileProblem.java
```diff
@@ -0,0 +1,371 @@
...
+    try (NetcdfFormatWriter writer = writerBuilder.build()) {
+      writer.write(variableName, values);
+    }
+  }
+
+  private static void writeNetcdfFileWithStructure(String filename, String structureName, String memberName,
+      DataType dataType, Array values) throws IOException, InvalidRangeException {
+    String structureDimensionName = "structureDimension";
+    String memberDimensionName = "memberDimension";
+    NetcdfFormatWriter.Builder writerBuilder =
```
haileyajohnson: Same here

cdm/gcdm/src/main/resources/logback.xml
haileyajohnson, line 1: I am confused why this is needed

cdm/gcdm/src/main/java/ucar/gcdm/GcdmConverter.java
tdrwenski: Yes, that is definitely nicer! I used one bidirectional map in 81bd7991b6

cdm/gcdm/build.gradle
tdrwenski: We do want logback for the tests (I think this is what we usually use). But we also need a logger for `runtime` so that we can get the gcdm server logs (put in `$buildDir/gcdm_logs/gcdm.log`, see below). I don't know which implementation is preferred for this?

lesserwhirls: `testRuntimeOnly` extends `runtimeOnly`, so we should be ok with only declaring the logging implementation as `runtimeOnly`. In terms of implementation, you can use whichever logging implementation you like (that's supported by slf4j), but the logging config files are logback specific, so you'd need to change those for whichever implementation you use (log4j2, log4j, etc.).

cdm/gcdm/src/test/java/ucar/gcdm/TestGcdmNetcdfFileProblem.java
tdrwenski: Yes, good idea-- fixed in 21acb5d93b

cdm/gcdm/src/test/java/ucar/gcdm/TestGcdmNetcdfFileProblem.java
tdrwenski: fixed in 21acb5d93b

cdm/gcdm/src/test/java/ucar/gcdm/TestGcdmNetcdfFileProblem.java
tdrwenski: Completely agree! I moved the tests into separate classes with more specific names in d3d75b5170. Some tests still have vague names but I didn't want to delete them if I didn't know what they were testing 😓 .

cdm/gcdm/build.gradle
tdrwenski: Removed `testRuntimeOnly` in c86d0a118f. Left `runtimeOnly` logger implementation as logback-classic, let me know if you prefer something else!

cdm/gcdm/src/main/resources/logback.xml
tdrwenski: This config file sets the log level of the gCDM server logs, whereas the logback-test.xml sets the log level of the tests and client code. I think it's nice to be able to configure both independently.

haileyajohnson (approved):

PR #1156, opened 2023-03-16
1D time - time=present should return the subset of time closest to the present 

mnlerman:
Adjust software behavior to match the [documentation](https://docs.unidata.ucar.edu/tds/current/userguide/ncss_grid.html#temporal-subsetting-and-valid-time-ranges)

`time=present ` should return the subset of time closest to the present 

## PR Checklist

- [x] Add labels
       until ready for review


cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoverageCoordAxis1D.java
```diff
@@ -396,7 +396,8 @@ protected Optional<CoverageCoordAxisBuilder> subsetBuilder(SubsetParams params)
 
       case Time:
         if (params.isTrue(SubsetParams.timePresent))
-          return Optional.of(helper.subsetLatest());
+          return Optional.of(helper.subsetClosest(CalendarDate.present()));
```
haileyajohnson, line 399: 🤦 this one just hurts

cdm/core/src/test/java/ucar/nc2/ft2/coverage/TestCoverageCoordAxis1D.java
```diff
@@ -0,0 +1,52 @@
...
+    SubsetParams presentSubsetParams = new SubsetParams();
+    presentSubsetParams.setTimePresent();
+
+    CoverageCoordAxis1D timeAxis = create1DTimeAxis();
+    Optional<CoverageCoordAxis> subsetTimeAxis = timeAxis.subset(presentSubsetParams);
+
+    assertThat(subsetTimeAxis.isPresent()).isTrue();
+
+    if (!subsetTimeAxis.isPresent()) {
+      return;
```
haileyajohnson: doesn't this not do anything if we made it through the first assert?

mnlerman: I added the condition to avoid NPEs on the later asserts since the subsetTimeAxis is optional

cdm/core/src/test/java/ucar/nc2/ft2/coverage/TestCoverageCoordAxis1D.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private CoverageCoordAxis1D create1DTimeAxis() {
+    CalendarDate refTime = testDate.subtract(CalendarPeriod.of(timeDelta, CalendarPeriod.Field.Day));
+    String timeUnit = "Day since " + refTime;
+
+    int valuesLen = 10;
+    double[] values = new double[valuesLen];
+    for (int i = 0; i < valuesLen; i++) {
+      values[i] = i + 1;
+    }
```
tdrwenski: Maybe simpler to just do
```suggestion
    double[] values = new double[] {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
```

cdm/core/src/test/java/ucar/nc2/ft2/coverage/TestCoverageCoordAxis1D.java
tdrwenski: I think if the optional is empty then the test will fail on line 26 and not reach this condition

haileyajohnson (approved):

cdm/core/src/test/java/ucar/nc2/ft2/coverage/TestCoverageCoordAxis1D.java
mnlerman: Ah yeah, that makes a lot of sense! Sorry Optionals are new to me :)

PR #1163, opened 2023-03-28
refactor implementation of scale/offset, unsigned conversion, and convert missing

haileyajohnson:


WeatherGod:
I'm currently investigating a performance issue in TDS serving up an NDVI dataset encoded with add_offset and scale_factor. I suspect some sort of memory leak or at best superfluous memory usage. I haven't looked through this PR yet, but do you suspect it might address those issues? Is it in a state where I could try it out myself and see if the problem is better or worse?

cdm/core/src/main/java/ucar/nc2/filter/UnsignedConversion.java
```diff
@@ -0,0 +1,69 @@
...
+    this.outType = outType;
+  }
+
+  public DataType getOutType() {
+    return this.outType;
+  }
+
+  public Number convertUnsigned(Number value) {
+    Number newVal = DataType.widenNumberIfNegative(value);
+    return DataType.widenNumberIfNegative(value);
```
WeatherGod: duplicate line of code?

cdm/core/src/test/java/ucar/nc2/filter/TestEnhancements.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Test
+  public void testCombinedEnhancements() throws IOException {
+    int[] expected = new int[] {1240, 1250, 1260, 1270, 1280, 0, 1278, 1260, 0, 0};
+    Variable v = ncd.findVariable("enhanceAll");
+    Array data = v.read();
+    assertThat((int[]) data.copyTo1DJavaArray()).isEqualTo(expected);
+  }
+}
```
WeatherGod, line 155: One test I'd be curious to see is if I take values that are from -1 to 1 and encode them as unsigned 8 bit integers (so, scale_factor = 0.008 and add_offset = -1), will the decode process give me back an array of floats rather than doubles. It isn't immediately clear to me that it would do that given that internally the scale and offset are stored as doubles.

cdm/core/src/main/java/ucar/nc2/filter/UnsignedConversion.java
haileyajohnson: Hi @WeatherGod , this PR is, as it says, a work in progress.
I have this duplicated atm so I can hit a breakpoint and see values here. I will get it cleaned up before it's ready to review. Thanks!

haileyajohnson:
> I'm currently investigating a performance issue in TDS serving up an NDVI dataset encoded with add_offset and scale_factor. I suspect some sort of memory leak or at best superfluous memory usage. I haven't looked through this PR yet, but do you suspect it might address those issues? Is it in a state where I could try it out myself and see if the problem is better or worse?

There's more changes to follow this PR that should fix some of the performance issues, I'd hold off on testing for a bit.

cdm/core/src/test/java/ucar/nc2/filter/TestFilters.java
```diff
@@ -96,6 +96,16 @@ public void testScaleOffset() throws IOException {
     // test decode
     byte[] decoded = filter.decode(encoded);
     assertThat(decoded).isEqualTo(input);
+
+    // test empty props
+    Map<String, Object> props2 = new HashMap<>();
```
tdrwenski: maybe names could be improved: `props2` -> `emptyProps`, `filter2` -> `emptyFilter`, `"fixedscaleoffset"` -> `"emptyScaleOffset"`

cdm/core/src/test/java/ucar/nc2/filter/TestEnhancements.java
```diff
@@ -0,0 +1,144 @@
...
+    data = v.read();
+    assertThat((float[]) data.copyTo1DJavaArray()).isEqualTo(expected);
+  }
+
+  @Test
+  public void testCombinedEnhancements() throws IOException {
+    double[] expected = new double[] {655320, Double.NaN, 655340, 655350, Double.NaN, 10, 20, 30, 40, 50};
+    Variable v = ncd.findVariable("enhanceAll");
+    Array data = v.read();
+    assertThat((double[]) data.copyTo1DJavaArray()).isEqualTo(expected);
```
tdrwenski, line 153: it looks like `enhanceAll` is a short, not a double

cdm/core/src/test/java/ucar/nc2/filter/TestEnhancements.java
```diff
@@ -0,0 +1,144 @@
...
+
+  private static final Short SIGNED_SCALED_MAX = -2;
+  private static final Short SIGNED_SCALED_FILL_VALUE = -4;
+
+  @ClassRule
+  public static final TemporaryFolder tempFolder = new TemporaryFolder();
+
+  @BeforeClass
+  public static void setUp() throws IOException, InvalidRangeException {
+    final int data_len = 10;
```
tdrwenski: maybe call this `dataLength`

cdm/core/src/main/java/ucar/nc2/dataset/EnhanceScaleMissingUnsigned.java
```diff
@@ -31,7 +31,7 @@
  * <li>Values will be {@link DataType#widenNumber widened}, which effectively reinterprets signed data as unsigned
  * data.</li>
  * <li>To accommodate the unsigned conversion, the variable's data type will be changed to the
- * {@link EnhanceScaleMissingUnsignedImpl#nextLarger(DataType) next larger type}.</li>
+ * {@link ucar.nc2.filter.FilterHelpers#nextLarger(DataType) next larger type}.</li>
```
tdrwenski, line 34: I see one more documentation link to `EnhanceScaleMissingUnsignedImpl#largestOf` on line 41 below

cdm/core/src/main/java/ucar/nc2/filter/ScaleOffset.java
```diff
@@ -83,27 +145,45 @@ public int getId() {
...
-    return arrayToBytes(out, dtype, dtypeOrder);
+    if (scale == DEFAULT_SCALE && offset == DEFAULT_OFFSET) {
+      return dataIn;
+    }
+    Array out = removeScaleOffset(FilterHelpers.bytesToArray(dataIn, astype, astypeOrder));
+    return FilterHelpers.arrayToBytes(out, dtype, dtypeOrder);
   }
 
-  private Array applyScaleOffset(Array in) {
+  // not used anywhere yet
```
tdrwenski: Maybe keep this `private` if its not being used outside this class

cdm/core/src/test/java/ucar/nc2/filter/TestEnhancements.java
```diff
@@ -0,0 +1,144 @@
...
+    builder.addDimension("dim", data_len);
+
+    Array signedData = Array.factory(DataType.SHORT, new int[] {data_len}, signedShorts);
+    // signed shorts
+    builder.addVariable("signedVar", DataType.SHORT, "dim");
+    // unsigned shorts
+    builder.addVariable("unsignedVar", DataType.SHORT, "dim").addAttribute(new Attribute(CDM.UNSIGNED, "true"));
+
+    // scaled and offset data
+    builder.addVariable("scaleOffsetVar", DataType.SHORT, "dim").addAttribute(new Attribute(CDM.SCALE_FACTOR, 10))
```
tdrwenski, line 53: maybe is good to have one test case of scale without offset or vice versa

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -609,172 +612,211 @@ public void removeCoordinateSystem(CoordinateSystem cs) {
...
   @Override
   public double getOffset() {
-    return scaleMissingUnsignedProxy.getOffset();
+    return scaleOffset != null ? scaleOffset.getOffset() : 0;
   }
 
   @Override
   public boolean hasMissing() {
-    return scaleMissingUnsignedProxy.hasMissing();
+    return convertMissing != null ? convertMissing.hasMissing() : false;
```
tdrwenski: This and several of the methods below that return booleans can be simplified like:
```suggestion
    return convertMissing != null && convertMissing.hasMissing()
```

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -609,172 +612,211 @@ public void removeCoordinateSystem(CoordinateSystem cs) {
...
+    if (Double.isNaN(val)) {
+      return true;
+    }
+    return convertMissing != null ? convertMissing.isMissing(val) : false;
   }
 
   @Override
   public boolean hasValidData() {
-    return scaleMissingUnsignedProxy.hasValidData();
+    return convertMissing != null ? convertMissing.hasMissingValue() : false;
```
tdrwenski:
```suggestion
    return convertMissing != null && convertMissing.hasMissingValue();
```

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -609,172 +612,211 @@ public void removeCoordinateSystem(CoordinateSystem cs) {
...
   }
 
   @Override
   public Number convertUnsigned(Number value) {
-    return scaleMissingUnsignedProxy.convertUnsigned(value);
+    return unsignedConversion != null ? unsignedConversion.convertUnsigned(value) : value;
+  }
+
+  public Number convertUnsigned(Number value, DataType dataType) {
+    return unsignedConversion != null ? unsignedConversion.convertUnsigned(value, dataType) : value;
```
tdrwenski, line 763: Not sure why you need this overload-- the `convertUnsigned(value, dataType)` being called here is the `static` version, maybe this is a mistake?

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -609,172 +612,211 @@ public void removeCoordinateSystem(CoordinateSystem cs) {
...
   @Override
   public double applyScaleOffset(Number value) {
-    return scaleMissingUnsignedProxy.applyScaleOffset(value);
+    return scaleOffset != null ? scaleOffset.removeScaleOffset(value) : value.doubleValue();
   }
 
   @Override
   public Array applyScaleOffset(Array data) {
-    return scaleMissingUnsignedProxy.applyScaleOffset(data);
+    return scaleOffset != null ? scaleOffset.removeScaleOffset(data) : data;
```
tdrwenski, line 754: should this be `applyScaleOffset`?

cdm/core/src/main/java/ucar/nc2/filter/ScaleOffset.java
```diff
@@ -166,26 +249,14 @@ private static ByteOrder parseByteOrder(String dtype, ByteOrder defaultOrder) {
...
-        return ULONG;
-      case LONG:
-      case ULONG:
-        return DOUBLE;
-      default:
-        return dataType;
-    }
+  public double applyScaleOffset(Number value) {
+    double convertedValue = value.doubleValue();
+    return Math.round((convertedValue - offset) * scale);
```
tdrwenski: Should this be
```suggestion
    return (convertedValue - offset) * scale;
```
since the `scale` is now a `double`?

cdm/core/src/test/java/ucar/nc2/filter/TestEnhancements.java
haileyajohnson: it starts as a short but after being widened to be unsigned and scaled, it's a double.

cdm/core/src/test/java/ucar/nc2/filter/TestFilters.java
haileyajohnson: done

cdm/core/src/main/java/ucar/nc2/filter/ScaleOffset.java
haileyajohnson: I think this comment was to remind me that zarr writes aren't implemented yet and this will be needed as part of that.

cdm/core/src/main/java/ucar/nc2/filter/ScaleOffset.java
haileyajohnson: ooops

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
haileyajohnson: I'm not sure - that's what it was before, but that was when we only needed to be able to remove the filter for reads, now we will need to use it for writes as well. Personally I think "apply" should mean write and "remove" should mean read, but 🤷 Maybe that's too confusing because of the way this class works (which doesn't do writes)

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
haileyajohnson: yeah this is kinda a complicated design, I'm not in love with it. Basically we need the instance method (which is used above) to convert the variable data itself, which will get the datatype from the class instance. But the static method is needed for other properties on the variable like scale and offset, which can be a different datatype than the var, and that needs to be passed in directly.

cdm/core/src/main/java/ucar/nc2/filter/ScaleOffset.java
tdrwenski: I see now that removing the round will give incorrect results for non-floating point types, as they will later get cast, so the floor will be taken. However, with the round you get incorrect results for floating point scale/offsets.

Maybe a solution could be
```suggestion
    if (astype == DOUBLE || astype == FLOAT) {
      return (convertedValue - offset) * scale;
    }
    return Math.round((convertedValue - offset) * scale);
```

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -609,172 +613,211 @@ public void removeCoordinateSystem(CoordinateSystem cs) {
...
   @Override
   public boolean hasMissingValue() {
-    return scaleMissingUnsignedProxy.hasMissingValue();
+    return convertMissing != null && convertMissing.hasMissingValue();
   }
 
   @Override
   public double[] getMissingValues() {
-    return scaleMissingUnsignedProxy.getMissingValues();
+    return convertMissing != null ? convertMissing.getMissingValues() : new double[] {0};
```
tdrwenski, line 684: if `convertMissing == null` shouldn't this be an empty array like
```suggestion
    return convertMissing != null ? convertMissing.getMissingValues() : new double[0];
```

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
tdrwenski: Okay, but since it's static maybe then we can do 
```suggestion
    return unsignedConversion != null ? UnsignedConversion.convertUnsigned(value, dataType) : value;
```

cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
```diff
@@ -98,7 +98,7 @@ public enum Enhance {
      * Convert unsigned values to signed values.
      * For {@link ucar.nc2.constants.CDM#UNSIGNED} variables, reinterpret the bit patterns of any
      * negative values as unsigned. The result will be positive values that must be stored in a
-     * {@link EnhanceScaleMissingUnsignedImpl#nextLarger larger data type}.
+     * {@link ucar.nc2.filter.FilterHelpers#nextLarger larger data type}.
```
tdrwenski, line 101: on line 107 below there is a doc link to `EnhanceScaleMissingUnsigned#isMissing` which is now deprecated

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -266,8 +259,17 @@ Array convert(Array data, Set<NetcdfDataset.Enhance> enhancements) {
...
-      return scaleMissingUnsignedProxy.convert(data, enhancements.contains(Enhance.ConvertUnsigned),
-          enhancements.contains(Enhance.ApplyScaleOffset), enhancements.contains(Enhance.ConvertMissing));
+      if (enhancements.contains(Enhance.ConvertUnsigned) && unsignedConversion != null) {
+        data = unsignedConversion.convertUnsigned(data);
+      }
+      if (enhancements.contains(Enhance.ApplyScaleOffset) && scaleOffset != null) {
+        data = scaleOffset.removeScaleOffset(data);
+      }
+      if (enhancements.contains(Enhance.ConvertMissing) && convertMissing != null
+          && (dataType == DataType.FLOAT || dataType == DataType.DOUBLE)) {
```
tdrwenski, line 269: maybe you could move the dataType checks `(dataType == DataType.FLOAT || dataType == DataType.DOUBLE)` to `createEnhancements`. That would make it more consistent with what's done for `ScaleOffset`.

cdm/core/src/main/java/ucar/nc2/filter/ConvertMissing.java
```diff
@@ -0,0 +1,262 @@
+package ucar.nc2.filter;
+
+import ucar.ma2.Array;
+import ucar.ma2.DataType;
+import ucar.ma2.IndexIterator;
+import ucar.nc2.Attribute;
+import ucar.nc2.constants.CDM;
+import ucar.nc2.constants.DataFormatType;
+import ucar.nc2.dataset.VariableDS;
+import ucar.nc2.iosp.netcdf3.N3iosp;
```
tdrwenski: unused imports 
```suggestion
import ucar.nc2.dataset.VariableDS;
```

cdm/core/src/test/java/ucar/nc2/filter/TestEnhancements.java
```diff
@@ -0,0 +1,153 @@
...
+        .addAttribute(Attribute.builder(CDM.VALID_RANGE).setValues(range).build())
+        .addAttribute(Attribute.builder(CDM.FILL_VALUE).setNumericValue(FILL_VALUE, true).build());
+
+    // unsigned, scaled/offset, and missing value
+    Array enhanceAllArray = Array.factory(DataType.SHORT, new int[] {dataLen}, signedShorts);
+    builder.addVariable("enhanceAll", DataType.SHORT, "dim").addAttribute(new Attribute(CDM.UNSIGNED, "true"))
+        .addAttribute(new Attribute(CDM.SCALE_FACTOR, 10.0)).addAttribute(new Attribute(CDM.ADD_OFFSET, 10))
+        .addAttribute(new Attribute(CDM.VALID_MAX, SIGNED_SCALED_MAX))
+        .addAttribute(new Attribute(CDM.FILL_VALUE, SIGNED_SCALED_FILL_VALUE));
+
```
tdrwenski, line 77: Maybe it's nice to test `CDM.MISSING_VALUE` as well

tdrwenski (changes requested):

haileyajohnson:
This should be g2g now :)

tdrwenski (approved):

PR #1167, opened 2023-04-10
Fix longitude subsetting when all longitudes are requested

### LINKED ISSUE #315: Problem getting hemispheric subset with NCSS
[Trying](https://github.com/Unidata/MetPy/discussions/2783) to download a hemispheric subset of GFS 0.5 degree data from Unidata's thredds server (looks like a 5.5 snapshot around 14 Sept 2022). The following requests:

https://thredds-dev.unidata.ucar.edu/thredds/ncss/grid/grib/NCEP/GFS/Global_0p5deg/GFS_Global_0p5deg_20221109_1200.grib2?var=Geopotential_height_isobaric&time=2022-11-09T18%3A50%3A37.350657&west=0&east=360&south=0&north=90&accept=netcdf4&addLatLon=True&vertCoord=25000

and

https://thredds-dev.unidata.ucar.edu/thredds/ncss/grid/grib/NCEP/GFS/Global_0p5deg/GFS_Global_0p5deg_20221109_1200.grib2?var=Geopotential_height_isobaric&time=2022-11-09T18%3A50%3A37.350657&west=-180&east=180&south=0&north=90&accept=netcdf4&addLatLon=True&vertCoord=25000

result in a return with a single longitude line of data. It's possible to work around this by e.g. changing 360 to 359.9. It would be really nice if instead the original queries worked like the user might expect (and did on TDS 4).

tdrwenski:
## Description of Changes

Fixes https://github.com/Unidata/tds/issues/315. When all longitudes are requested but not all latitudes (e.g. lon from [-180, 180]), and the longitude axis data is say {0... 360}, then the desired longitudes were getting normalized to be in the range [0, 360]. This causes the requested longitudes to be [180, 180] which results in only returning one point instead of all longitudes.

This PR uses the `allLongitude` boolean in the requested lat/lon box to fix this situation and adds a unit test for this.

Also added tests for NCSS in TDS PR: https://github.com/Unidata/tds/pull/354

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
+++ b/cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
@@ -330,14 +330,15 @@ public class HorizCoordSys {
   }
 
   // here's where to deal with crossing seam
-  private Optional<CoverageCoordAxis> subsetLon(LatLonRect llbb, int stride) {
-    double wantMin = LatLonPoints.lonNormalFrom(llbb.getLonMin(), lonAxis.getStartValue());
-    double wantMax = LatLonPoints.lonNormalFrom(llbb.getLonMax(), lonAxis.getStartValue());
+  private Optional<CoverageCoordAxis> subsetLon(LatLonRect latLonBoundingBox, int stride) {
+    double wantMin = LatLonPoints.lonNormalFrom(latLonBoundingBox.getLonMin(), lonAxis.getStartValue());
+    double wantMax = LatLonPoints.lonNormalFrom(latLonBoundingBox.getLonMax(), lonAxis.getStartValue());
     double start = lonAxis.getStartValue();
     double end = lonAxis.getEndValue();
 
     // use MAMath.MinMax as a container for two values, min and max
-    List<MAMath.MinMax> lonIntvs = subsetLonIntervals(wantMin, wantMax, start, end, llbb.containsAllLongitude());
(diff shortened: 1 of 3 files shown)


cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
```diff
@@ -337,7 +337,7 @@ private Optional<CoverageCoordAxis> subsetLon(LatLonRect llbb, int stride) {
     double end = lonAxis.getEndValue();
 
     // use MAMath.MinMax as a container for two values, min and max
-    List<MAMath.MinMax> lonIntvs = subsetLonIntervals(wantMin, wantMax, start, end);
+    List<MAMath.MinMax> lonIntvs = subsetLonIntervals(wantMin, wantMax, start, end, llbb.containsAllLongitude());
```
haileyajohnson: Can you add a comment somewhere that llbb = latlonboundingbox? I'm trying to get in the habit of labeling these things since we keep ending up going ??? about them haha

haileyajohnson (approved):

cdm/core/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
tdrwenski: Yes, good point! I just updated the variable name.

PR #1170, opened 2023-04-14
Fix MFileS3::exists to work for "directories"

tdrwenski:
## Description of Changes

Part of the fixes needed for S3 datasetScans.

The current `MFileS3::exists` only works for Objects, and uses a `headObject` request. This PR would extend this to work for buckets (need to do a `headBucket`) and "directories" ( by doing a `listObjects` and checking it has objects).

Note that "Directories" are defined in our code as URIs that have a delimiter fragment and also end with that delimiter unless it's a bucket (e.g. `cdms3:myBucket#delimiter=/` or `cdms3:myBucket?myKey/#delimiter=/`). 

I wonder if `cdms3:myBucket` should also be considered a directory?

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/s3/src/main/java/thredds/inventory/s3/MFileS3.java
+++ b/cdm/s3/src/main/java/thredds/inventory/s3/MFileS3.java
@@ -315,7 +315,7 @@ public class MFileS3 implements MFile {
     try {
       final ListObjectsV2Response listObjects =
           client.listObjectsV2(ListObjectsV2Request.builder().bucket(cdmS3Uri.getBucket()).prefix(key).build());
-      return !listObjects.contents().isEmpty();
+      return listObjects.sdkHttpResponse().isSuccessful() && !listObjects.contents().isEmpty();
     } catch (NoSuchBucketException e) {
       return false;
     }
@@ -323,8 +323,8 @@ public class MFileS3 implements MFile {
 
   private boolean bucketExists() {
     try {
-      getHeadBucketResponse();
-      return true;
+      final HeadBucketResponse response = getHeadBucketResponse();
+      return response != null && response.sdkHttpResponse().isSuccessful();
     } catch (NoSuchBucketException e) {
       return false;
     }
@@ -332,8 +332,8 @@ public class MFileS3 implements MFile {
 
   private boolean objectExists() {
     try {
-      headObjectResponse.get();
-      return true;
+      final HeadObjectResponse response = headObjectResponse.get();
+      return response != null && response.sdkHttpResponse().isSuccessful();
     } catch (NoSuchKeyException e) {
       return false;
     }
--- a/cdm/s3/src/main/java/thredds/inventory/s3/MFileS3.java
+++ b/cdm/s3/src/main/java/thredds/inventory/s3/MFileS3.java
@@ -104,6 +104,19 @@ public class MFileS3 implements MFile {
   }
 
   @Nullable
+  private HeadBucketResponse getHeadBucketResponse() {
+    HeadBucketResponse response = null;
+    S3Client client = getClient();
+
+    if (client != null) {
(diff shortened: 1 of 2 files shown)


lesserwhirls:
> I wonder if `cdms3:myBucket` should also be considered a directory?

That's a good question. I had thought of it like this in terms of a mapping to file systems. 

`cdms3:myBucket#delimiter=/` was essentially saying that the objects in bucket the were laid out such that they could be interpreted as filesystem paths, and the bucket was considered to be the filesystem root "/". Performing a list-objects-v2 request against the bucket with a delimiter of "/" would then be like doing `ls /`. In the json return from AWS, `"Contents": [...]` would then, essentially, contain a list of the "files" directly under "/", while `"CommonPrefixes": [..]` would contain a list of directories directly under "/". So in that sense, everything in the bucket is accessible as if it were filesystem paths without too many surprises. We can then browse the bucket using a combination of delimiter="/" and prefix (selected from the common prefixes at a given "level" of the bucket).

Now let's say the objects in bucket the are not laid out such that they could be interpreted as filesystem paths, but you'd still like to treat the bucket as if it was a filesystem (so you could do things like a "scan" operation of some sort...maybe a datasetScan?). In this case there would be one top level directory with, potentially, a ton of "files", and no subdirectories. The json return from AWS would have `"Contents": [...]` and that's it. The root directory would exist and would map to the `cdms3:myBucket`, so maybe that should be included as a directory "existing"?

Looking further down the road a bit, treating the bucket as a directory that exists would enable a user to trigger filesystem like operations that would bring their system to a scratching halt, since listing objects in a bucket without a delimiter and/or prefix can become very expensive (time-wise). For example, using the aws cli, compare `aws s3api list-objects-v2 --region us-east-1 --bucket noaa-goes18 --delimiter="/"` with `aws s3api list-objects-v2 --region us-east-1 --bucket noaa-goes18`. Just something to think about.

tdrwenski:
That makes sense-- way too many objects would be listed if a bucket was considered a directory without a delimiter. Thanks @lesserwhirls  :)

cdm/s3/src/main/java/thredds/inventory/s3/MFileS3.java
```diff
@@ -285,13 +298,44 @@ public boolean exists() {
...
+    } catch (NoSuchBucketException e) {
+      return false;
+    }
+  }
+
+  private boolean bucketExists() {
+    try {
+      getHeadBucketResponse();
+      return true;
+    } catch (NoSuchBucketException e) {
```
haileyajohnson, line 328: Sorry if this is just my bronchitis brain, but does this method throw that error? I thought it just returns null.

tdrwenski: Yes it does-- the `getHeadBucketResponse` does a `headBucket` which has this signature: 
```
default HeadBucketResponse headBucket(HeadBucketRequest headBucketRequest) throws NoSuchBucketException, AwsServiceException,
            SdkClientException, S3Exception
```
But the `NoSuchBucketException` is a `RuntimeException` so not declared for `getHeadBucketResponse` so it's not obvious. The test `shouldReturnFalseForNonExistentBucket` would fail without this catch statement.

tdrwenski: but you are right it could also return `null` if there is no client so I should probably check that. And maybe we should also test the response code? Maybe there could be a bucket you don't have access to so it doesn't throw `NoSuchBucketExceptions`?

haileyajohnson: I think at least checking for null would be good

tdrwenski: Added null checks and response code checks in bf0a581e759eed0c7031cfe178a284cd8c4f596a

haileyajohnson (approved):

PR #1172, opened 2023-04-26
Change MFiles.create to be Nonnull

tdrwenski:
## Description of Changes

Currently, `MFiles.create` claims that it both creates an `MFile` using the `MFileProviders` or by using `MFileOS` and also checks if the file exists. However, the existence check only applies in the case of the `MFileOS` and not for instance `MFileS3` or `MFileZip`. Note that we now have a separate method for checking existence (`MFile::exists`).

This PR would remove checking for existence in the `create` function. I think this change makes the code nicer since the `create` wasn't really checking existence for all types of `MFiles`. It also avoids null checks when using `create`. It makes testing simpler, for instance, I want to make tests that use `MFile.create` to test something that doesn't necessarily require files to exist.

I do not believe this counts as a breaking change, though it does change the documented behavior of this function. I added a second variant `createIfExisting` that returns null if `exists` is false (old behavior for `create` when its a `MFileOS`). Can change the names if you prefer!

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/core/src/main/java/thredds/inventory/MFileProvider.java
+++ b/cdm/core/src/main/java/thredds/inventory/MFileProvider.java
@@ -6,7 +6,7 @@
 package thredds.inventory;
 
 import java.io.IOException;
-import javax.annotation.Nullable;
+import javax.annotation.Nonnull;
 
 /**
  * A Service Provider of {@link thredds.inventory.MFile}.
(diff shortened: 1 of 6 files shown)


cdm/core/src/main/java/thredds/inventory/MFileProvider.java
```diff
@@ -22,12 +22,11 @@ default boolean canProvide(String location) {
   }
 
   /**
-   * Create an {@link thredds.inventory.MFile} from a given location if it exists and is readable, otherwise return
-   * null.
+   * Create an {@link thredds.inventory.MFile} from a given location, the file may or may not exist
    *
    * @param location location of a file or .
    * @return {@link thredds.inventory.MFile}
    */
```
haileyajohnson, line 29: I actually don't know if this is a breaking change haha. I think that it's not because it's going nullable to nonnull, but nonnull to nullable would be? Because in this case, users might just be checking for null when they don't have to...does that sound right?

haileyajohnson (approved):

Issue #1188, opened 2023-05-23
Remove groovy/spock dependencies


tdrwenski:
### Overview
There are currently 18 `.groovy` files that have Spock tests in NetCDF-java. These should be converted to JUnit tests and then the dependencies on groovy and spock should be removed from the gradle build files. Some tests may need to to be altered because you can test private methods with Spock but not with JUnit.

### Files
To convert:
```
./cdm/core/src/test/groovy/ucar/nc2/ft2/coverage/HorizCoordSysCrossSeamBoundarySpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/ft2/coverage/writer/CFGridCoverageWriter2Spec.groovy
./cdm/core/src/test/groovy/ucar/nc2/ft2/coverage/HorizCoordSysGisTextBoundarySpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/util/cache/ReacquireClosedDatasetSpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/ncml/LogicalViewsSpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/ncml/NcMLWriterSpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/ncml/CacheAggregationsSpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/iosp/netcdf3/N3iospSpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/ft/point/FlattenedDatasetPointCollectionSpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/ft/point/PointIteratorFilteredSpec.groovy
./cdm/core/src/test/groovy/ucar/nc2/ft/point/remote/PointStreamSpec.groovy
./netcdf4/src/unloadedTest/groovy/ucar/nc2/jni/netcdf/UnloadedNc4IospSpec.groovy
./netcdf4/src/test/groovy/ucar/nc2/jni/netcdf/Nc4IospSpec.groovy
./netcdf4/src/test/groovy/ucar/nc2/jni/netcdf/Nc4IospMiscSpec.groovy
```

These tests are testing no longer used legacy code and so can just be removed:
```
./legacy/src/test/groovy/thredds/crawlabledataset/s3/ThreddsS3ClientImplSpec.groovy
./legacy/src/test/groovy/thredds/crawlabledataset/s3/CachingThreddsS3ClientSpec.groovy
[...]

PR #1189, opened 2023-05-30
Convert from groovy to junit

cssjessica:
## Description of Changes

The following spock groovy test were converted to Junit: PointStreamSpec and Nc4IospSpec from Issue https://github.com/Unidata/netcdf-java/issues/1188.

## PR Checklist

- [ ] Add labels
       until ready for review


cdm/core/src/test/java/ucar/nc2/ft/point/remote/TestPointStream.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import ucar.nc2.constants.FeatureType;
+import ucar.nc2.ft.FeatureDatasetFactoryManager;
+import ucar.nc2.ft.FeatureDatasetPoint;
+import ucar.nc2.ft.NoFactoryFoundException;
+import ucar.nc2.ft.PointFeatureCollection;
+import ucar.nc2.ft.point.FlattenedDatasetPointCollection;
+import ucar.nc2.ft.point.PointTestUtil;
+import ucar.nc2.ft.point.remote.PointCollectionStreamLocal;
```
haileyajohnson: These two imports are unused (PointCollectionStreamLocal and PointStream)

cdm/core/src/test/java/ucar/nc2/ft/point/remote/TestPointStream.java
```diff
@@ -0,0 +1,72 @@
...
+    this.location = location;
+  }
+
+  @Test
+  public void roundTrip() throws IOException, NoFactoryFoundException {
+
+
+    File outFile = temporaryFolder.newFile();
+    try (FeatureDatasetPoint fdPoint =
+        (FeatureDatasetPoint) FeatureDatasetFactoryManager.open(FeatureType.ANY_POINT, location, null)) {
```
haileyajohnson: This method is deprecated, the new method signature takes and error log formatter (which can just be null for the sake of using the current API). This test also then won't throw the `NoFactoryFoundException`

cdm/core/src/test/java/ucar/nc2/ft/point/remote/TestPointStream.java
```diff
@@ -0,0 +1,72 @@
...
+
+  String location;
+
+  public TestPointStream(String location) {
+    this.location = location;
+  }
+
+  @Test
+  public void roundTrip() throws IOException, NoFactoryFoundException {
+
```
haileyajohnson: Delete extra whitespace :)

cdm/core/src/test/java/ucar/nc2/ft/point/remote/TestPointStream.java
```diff
@@ -0,0 +1,72 @@
...
+      PointFeatureCollection origPointCol = new FlattenedDatasetPointCollection(fdPoint);
+      PointStream.write(origPointCol, outFile);
+      PointFeatureCollection roundTrippedPointCol = new PointCollectionStreamLocal(outFile);
+
+      assertThat(PointTestUtil.equals(origPointCol, roundTrippedPointCol)).isTrue();
+    }
+
+  }
+
```
haileyajohnson: Delete extra whitespace :)

netcdf4/src/test/groovy/ucar/nc2/jni/netcdf/Nc4IospSpec.groovy
```diff
@@ -1,46 +0,0 @@
...
- * @since 2016-12-27
- */
-class Nc4IospSpec extends Specification {
-    private static final Logger logger = LoggerFactory.getLogger(Nc4IospSpec)
-    
-    def setup() {
-        // Ignore this class's tests if NetCDF-4 isn't present.
-        // We're using setup() because it shows these tests as being ignored.
-        // setupSpec() shows them as *non-existent*, which is not what we want.
-        Assume.assumeTrue("NetCDF-4 C library not present.", Nc4Iosp.isClibraryPresent());
```
haileyajohnson, line 22: I think we still need to check that the C library is present for these tests. You can follow what this test does: https://github.com/Unidata/tds/blob/13017bd6dfd9f5ae270bf2e524d8d7340592742b/tds/src/test/java/thredds/server/ncss/view/dsg/DsgSubsetWriterTest.java#L47

haileyajohnson (changes requested):

netcdf4/src/test/groovy/ucar/nc2/jni/netcdf/Nc4IospSpec.groovy
tdrwenski: If I understand the comment correctly they wanted the tests to be ignored if the C Library isn't present. But these tests just test that no exception is thrown when a method is called, which should be the case whether or not the library exists, so I don't think that assumption is really needed

haileyajohnson: wait, what are these even meant to be testing?

tdrwenski: My understanding was that at some point if you didn't have the netcdf-c library present then calling these functions caused a null pointer exception. Then they made these tests and added a null check to fix that. So they only check that there is no null pointer exception thrown, but yeah not the most useful tests of all time 😆

haileyajohnson: lol well now they don't do anything cuz the functions return right away 🤷

tdrwenski: yeah it's not a super useful test... do you still want to check for the c library or leave as is?

cdm/core/src/test/java/ucar/nc2/ft/point/remote/TestPointStream.java
cssjessica: The method was updated.

netcdf4/src/test/groovy/ucar/nc2/jni/netcdf/Nc4IospSpec.groovy
cssjessica: We ended up adding the check for the C library

haileyajohnson (approved):

Issue #1206, opened 2023-07-03
Add Standardizer preprocessor to netcdf dataset enhancements


cssjessica:
[Standardization](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler.transform) is a common preprocessing step for AI/ML inputs. So, the goal is to have a raw and a preprocessed version of datasets in the TDS Catalog. The Standardizer preprocessing can be achieved through NcML as

```
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2" enhance="all">
  <variable name="scaledVar">
    <attribute name="standardize"/>
  </variable>
</netcdf>
```

- [x] Compute the mean and standard deviation for scaling
- [x] Perform scaling using Scale Offset
- [x] Add to set of enhancements and apply when correct attribute exists
- [x] Test with a dataset on thredds-test server
- [x] Add notebooks for visualization
- [x] Test performance

haileyajohnson:
This is great!
I'd like to talk about the name of the enhancement before it's too far along. To me "scaler" is a bit confusing because it's so similar to "scalar". So maybe just `Standardize` or `Normalize`?

cssjessica:
Yes! @tdrwenski also commented about it being a bit confusing. I'll change to `Normalize` then.

tdrwenski:
I think "normalize" makes it sound like it is rescaled to unit length so I think I prefer "standardize". So would we call the class that performs the operation a `Standardizer` (since its good to use a noun for the class name)?

cssjessica:
That's true! With standardization, the values don't have a bounding range

haileyajohnson:
awesome, thanks @cssjessica !

PR #1210, opened 2023-07-07
Standardizer

cssjessica:
## Description of Changes

Add Standardizer as an enhancement. For more background, see issue #1206 

## PR Checklist

- [x] Add labels
       until ready for review


cdm/core/src/test/java/ucar/nc2/ncml/TestEnhanceStandardizer.java
```diff
@@ -0,0 +1,52 @@
...
+      assertThat(nearlyEquals(dataDoubles, DATA_DOUBLES)).isTrue();
+
+      Variable floatVar = ncfile.findVariable("floatvar");
+      assertThat((Object) floatVar).isNotNull();
+      assertThat(floatVar.getDataType()).isEqualTo(DataType.FLOAT);
+      assertThat(doubleVar.attributes().hasAttribute("standardize")).isTrue();
+      Array dataFloats = doubleVar.read();
+      assertThat(nearlyEquals(dataFloats, DATA_FLOATS)).isTrue();
+
+      Variable intVar = ncfile.findVariable("intvar");
```
tdrwenski, line 43: maybe add a small comment here that the standardizer is not applied to ints so we expect the original data to be returned

cdm/core/src/main/java/ucar/nc2/filter/Standardizer.java
```diff
@@ -0,0 +1,80 @@
...
+    this.stdDev = stdDev;
+    Map<String, Object> props = new HashMap<>();
+    props.put("offset", mean);
+    props.put("scale", 1 / stdDev);
+    props.put("dtype", type);
+    scaleOffset = new ScaleOffset(props);
+  }
+
+  public Standardizer(Array arr, DataType type) {
+    this(calculateMean(arr), calculateStandardDeviation(arr), type);
```
tdrwenski: One optimization we could do (okay with me to wait on this til later though) is to only loop over the data once and get both the mean and standard deviation at the same time

cdm/core/src/test/java/ucar/nc2/ncml/TestEnhance.java
```diff
@@ -76,7 +76,7 @@ public void testStandaloneEnhance() throws IOException {
 
   @Test
   public void testStandaloneEnhanceDataset() throws IOException {
-    try (NetcdfFile ncfile = NetcdfDatasets.openDataset(dataDir + "testStandaloneNoEnhance.ncml", true, null)) {
+    try (NetcdfFile ncfile = NetcdfDatasets.openDataset(dataDir + "testStandaloneEnhance.ncml", true, null)) {
```
tdrwenski: Maybe @haileyajohnson can weigh in, but I think maybe this was actually intended to use `testStandaloneNoEnhance.ncml`, since the one below tests `testStandaloneEnhance.ncml`. I think this tests an enhanced dataset even with the "NoEnhance" file because of the `true` in the `openDataset`

cdm/core/src/main/java/ucar/nc2/filter/Standardizer.java
```diff
@@ -0,0 +1,80 @@
...
+import ucar.ma2.Array;
+import ucar.ma2.DataType;
+import ucar.ma2.IndexIterator;
+import ucar.nc2.dataset.VariableDS;
+
+public class Standardizer {
+
+  private final ScaleOffset scaleOffset;
+  private static final String name = "Standardizer";
+  private static final int id = -1;
```
tdrwenski:
```suggestion
```
We don't use these anymore now that the `Standardizer` isn't a `Filter`

cssjessica: Both lines were removed.

cdm/core/src/test/java/ucar/nc2/ncml/TestEnhanceStandardizer.java
cssjessica: Added the small comment for the expected outcome.

cdm/core/src/test/java/ucar/nc2/ncml/TestEnhance.java
cssjessica: Changed back to the original `testStandaloneNoEnhance.ncml`.

cdm/core/src/main/java/ucar/nc2/filter/Standardizer.java
cssjessica: Added a helper function and the constructor of Standardize was refactored, so it should now loop only once through the data for the calculations.

haileyajohnson (approved):

PR #1211, opened 2023-07-08
Fix checksum handling in DAP4.

DennisHeimbigner:
## Description of Changes
The DAP4 specification says how to properly handle checksums, but this was improperly implemented in netcdf-java. This PR makes some changes to handle checksums.

WARNING: The merging of this PR needs to be synchronized with the following netcdf-java PR: XXXX.
So after both PRs are reviewed and approved, both need to be merged one right after the other.

Note also that their are still unresolved ambiguities about checksums in the DAP4 specification that need resolution. See [here](https://github.com/OPENDAP/dap4-specification/discussions/6) for example. So at the moment, accessing non-thredds servers via DAP4 requires special treatment.

## PR Checklist

- [X] Add labels
- [X] Open as a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) until ready for review
- [X] Make sure GitHub tests pass
- [X] Make sure Jenkins tests pass

## DAP4 Changes
1. Properly compute and apply checksums in the DAP and DMR responses.
2. Fix tests that require the _DAP4_Checksum_CRC32 attribute.

## Non-DAP4 Changes
The only controversial change is that to ucar.ma2.Index, about line 573. The change is as follows:
````
    int prefixrank = (hasvlen ? rank: rank - 1);
to
    int prefixrank = (hasvlen ? rank - 1 : rank);
````
The original line is clearly an error under the assumption that "*" dimensions are always the last dimension for an array.  I was worried about this causing follow-on failures, but not surprisingly because vlen is so screwed up, no follow-on failures have appeared.

### DIFF
--- a/dap4/src/main/java/dap4/core/ce/CEConstraint.java
+++ b/dap4/src/main/java/dap4/core/ce/CEConstraint.java
@@ -334,7 +334,7 @@ public class CEConstraint {
     Segment seg = findSegment(var);
     if (seg != null)
       slices = seg.slices;
-    if (slices == null)
+    if (slices == null || slices.size() == 0)
       slices = Universal.universalSlices(var);
     return slices;
   }
--- a/dap4/src/main/java/dap4/core/ce/CEConstraint.java
+++ b/dap4/src/main/java/dap4/core/ce/CEConstraint.java
@@ -334,7 +334,7 @@ public class CEConstraint {
     Segment seg = findSegment(var);
     if (seg != null)
       slices = seg.slices;
-    if (slices == null || slices.isEmpty() == 0)
+    if (slices == null || slices.isEmpty())
       slices = Universal.universalSlices(var);
     return slices;
   }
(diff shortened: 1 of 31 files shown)

### CODE REVIEW

File: dap4/src/main/java/dap4/core/ce/CEConstraint.java [hunk 1]
```diff
@@ -334,7 +334,7 @@ public List<Slice> getConstrainedSlices(DapVariable var) throws DapException {
     Segment seg = findSegment(var);
     if (seg != null)
       slices = seg.slices;
-    if (slices == null)
+    if (slices == null || slices.size() == 0)
```
haileyajohnson: `slices.isEmpty()`?

haileyajohnson (approved):

dap4/src/main/java/dap4/core/ce/CEConstraint.java
DennisHeimbigner: Hah! I am imprinted with my C idoms. Will change.

DennisHeimbigner:
Fixed,

PR #1212, opened 2023-07-11
fix up some minor ncss issues

haileyajohnson:
- don't throw if a variable doesn't intersect the requested time
- refactor request too large exception so the TDS can recognize it more explicitly

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoordAxisHelper.java
+++ b/cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoordAxisHelper.java
@@ -11,7 +11,7 @@ import ucar.ma2.RangeIterator;
 import ucar.nc2.constants.AxisType;
 import ucar.nc2.time.CalendarDate;
 import ucar.nc2.time.CalendarDateRange;
-import ucar.nc2.util.Misc;
+
 import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


haileyajohnson:
I've tested this manually but I could probably add a unit test

cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoordAxisHelper.java
```diff
@@ -25,6 +25,8 @@
 class CoordAxisHelper {
   private final CoverageCoordAxis1D axis;
 
+  private static final int MUTIPLE_HITS = -2;
```
tdrwenski: spelling:
```suggestion
  private static final int MULTIPLE_HITS = -2;
```

cdm/core/src/main/java/ucar/nc2/ft2/coverage/CoordAxisHelper.java
```diff
@@ -204,13 +206,14 @@ else if (upperIndex > axis.getCoordEdgeLast())
   // LOOK not using bounded
```
tdrwenski: maybe this comment can go now that we are using `bounded`
```suggestion
```

tdrwenski (approved):

PR #1217, opened 2023-07-19
Normalizer

cssjessica:
## Description of Changes

Normalization is a scaling technique in which values range between 0 and 1. It is a common preprocessing step for AI/ML alongside Standardization #1206 . This PR will include a preprocessed version for datasets in the TDS Catalog. The Normalizer preprocessing can be achieved through NcML as

```
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2" enhance="all">
  <variable name="scaledVar">
    <attribute name="normalize"/>
  </variable>
</netcdf>
```

## PR Checklist

- [x] Add labels
       until ready for review


cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -860,6 +865,10 @@ private void createEnhancements() {
     if (standardizerAtt != null && this.enhanceMode.contains(Enhance.ApplyStandardizer) && dataType.isFloatingPoint()) {
       this.standardizer = Standardizer.createFromVariable(this);
     }
+    Attribute normalizerAtt = findAttribute(CDM.NORMALIZE);
+    if (normalizerAtt != null && this.enhanceMode.contains(Enhance.ApplyStandardizer) && dataType.isFloatingPoint()) {
```
tdrwenski: Think this should be
```suggestion
    if (normalizerAtt != null && this.enhanceMode.contains(Enhance.ApplyNormalizer) && dataType.isFloatingPoint()) {
```

cdm/core/src/test/java/ucar/nc2/filter/TestNormalizer.java
```diff
@@ -0,0 +1,78 @@
...
+    assertThat(calcRange).isWithin(TOLERANCE).of(DATA_RANGE);
+  }
+
+  @Test
+  public void testCalculateRangeFloat() {
+    Normalizer filter = Normalizer.createFromArray(DATA_FLOATS, DataType.FLOAT);
+    double calcRange = filter.getRange();
+    assertThat(calcRange).isWithin(TOLERANCE).of(DATA_RANGE);
+  }
+
```
tdrwenski, line 87: What happens if all the values are the same so that the range is zero?

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
cssjessica: Absolutely!

cdm/core/src/test/java/ucar/nc2/filter/TestNormalizer.java
cssjessica: Great point! All the values receive `NaN` and the same goes for `standardize` .

tdrwenski (approved):

PR #1233, opened 2023-09-15
Fix projection units

tdrwenski:
## Description of Changes

Some NCSS requests are failing for satellite data (geostationary projection) when the x/y axes have units micro radians. This PR fixes this by moving the unit handling out of the projection class, into the `HorizCoordSys` class that is used by coverages. This PR contains:
- Add test
- Generalize unit handling in `HorizCoordSys` using the `SimpleUnits` to work for more than just km/m, given that a projection specifies its default unit.
- Remove/deprecate code in `Geostationary` related to unit handling.
- Add default unit to `Geostationary` and `LambertConformal` classes.


tdrwenski:
Jenkins tests passing: https://jenkins-aws.unidata.ucar.edu/view/Users/job/tara-netcdf-java/25/

haileyajohnson (approved):

PR #1238, opened 2023-09-22
Fix disk cache cleanup sort errors caused by files being altered during sorting

okey:
## Description of Changes

Java versions newer than 6 check that the sort invariant is maintained. If a cached file is modified during sort, then the modified time changes, violating the invariant and causing an exception to be thrown. Previously, this would have silently caused the resulting sort to be incorrect.

Neither of these results are ideal, but throwing an exception causes the cache cleanup to fail, which can lead to unexpectedly high cache usage.

A quick and nasty hack to revert to Java 6 behaviour is to insert `System.setProperty("java.util.Arrays.useLegacyMergeSort", "true");` somewhere, but it is better to fix the sort completely.

This PR caches the file modification times and looks them up during the sort.

## PR Checklist

- [ ] ~Link to any issues that the PR addresses~
- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/util/DiskCache.java
+++ b/cdm/core/src/main/java/ucar/nc2/util/DiskCache.java
@@ -348,7 +348,8 @@ public class DiskCache {
     }
 
     public int compare(File f1, File f2) {
-      return lastModified.getOrDefault(f2, DateTime.now().getMillis()).compareTo(lastModified.getOrDefault(f1, DateTime.now().getMillis()));
+      return lastModified.getOrDefault(f2, DateTime.now().getMillis())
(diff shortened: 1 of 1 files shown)


haileyajohnson:
Thanks for the contribution, @okey! I refactored your solution a bit to still plug in a comparator, but one that uses a static `lastModified` property. As for `cleanCache(long maxBytes, Comparator<File> fileComparator, StringBuilder sbuff)`, I don't think we want to deprecate it since there are still reasons users may want to sub their own `Comparator` to sort the files in their cache.

okey:
@haileyajohnson Thanks for the refactor, if you're happy with using a static property then I think it's a good solution.
~Do you want to fix the test failures or shall I?~

tdrwenski (approved):

PR #1247, opened 2023-10-10
fix scale offset applied twice

haileyajohnson:
fix scale offset applied twice

remove extra calls to convert for performance

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
@@ -932,6 +932,10 @@ public class NetcdfDataset extends ucar.nc2.NetcdfFile {
     return ImmutableList.copyOf(coordAxes);
   }
 
+  public void setCoordinateAxes(List<CoordinateAxis> axes) {
+    this.coordAxes = axes;
+  }
+
   /**
    * Clear Coordinate System metadata, to allow them to be redone
    * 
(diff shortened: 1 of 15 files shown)


cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
```diff
@@ -1344,12 +1344,16 @@ public void addCoordinateTransform(CoordinateTransform ct) {
   public CoordinateAxis addCoordinateAxis(VariableDS v) {
     if (v == null)
       return null;
+
+    final List<CoordinateAxis> coordCopy = new ArrayList<>(coordAxes);
```
tdrwenski, line 1348: why is this copy of the list needed?

tdrwenski (commented):
Great job getting this done! 🎉  Do you think it's easy to add a test for the scale/offset applied twice to time axes bug? Some of this code may be hard to unit test but we could always add one to the TDS/ NCSS if that's easier.

cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
haileyajohnson: It's kinda unrelated to the other changes, but this method throws an error if you try to use it as is because `coordAxes` is immutable. Was just something I bumped into while testing stuff out though

haileyajohnson:
The scale/offset applied twice by nccs and the TDS bug is being tested already by all the tests that check that the scale and offset properties are removed, but I was also planning on putting a test in the TDS. The thing with the time axes calling `read` and `convert` over and over is kinda hard to test because it doesn't actually change any results, so for me to test it I had to manually count how many times that code was hit...

tdrwenski (approved):

PR #1248, opened 2023-10-13
[5.5.3]: Fix GRIB IOSP time coords from intervals, for 2-D time datasets

ethanrd:
## Description of Changes

Fix generation of time coordinate variables from time intervals for 2-D time datasets. Ensure resulting time coordinate variables are strictly monotonically increasing. Handle when time intervals are irregular and overlapping.

See issue #1060.

## PR Checklist

- [ ] Add labels
       until ready for review


grib/src/main/java/ucar/nc2/grib/collection/GribTimeCoordIntervalUtils.java
```diff
@@ -0,0 +1,75 @@
...
+
+public class GribTimeCoordIntervalUtils {
+  /**
+   * Default location in a time coordinate interval to use for the corresponding time coordinate value
+   * when generating a time coordinate variable.
+   */
+  enum TimeCoordValueDefaultLocationInIntervals {
+    endPoint, midPoint, startingPoint
+  }
+
```
tdrwenski: I don't believe this is being used
```suggestion
```

grib/src/main/java/ucar/nc2/grib/collection/GribTimeCoordIntervalUtils.java
```diff
@@ -0,0 +1,75 @@
+package ucar.nc2.grib.collection;
+
+import ucar.nc2.grib.coord.TimeCoordIntvValue;
+
+import java.util.List;
+
+public class GribTimeCoordIntervalUtils {
```
tdrwenski: Since this is only being used in this package, we could make this package private
```suggestion
class GribTimeCoordIntervalUtils {
```

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
```diff
@@ -94,4 +94,66 @@ private void checkTimeCoordinateVariable1D_IsStrictlyMonotonicallyIncreasing(int
...
+  }
+
+  private void checkTimeCoord2D_fromIntervals_isStrictlyMonotonicallyIncreasing(String testfile, String varName)
+      throws IOException {
+    try (NetcdfFile nc = NetcdfFiles.open(testfile)) {
+      Variable dataVar = nc.findVariable(varName);
+      Assert.assertNotNull(dataVar);
+
+      Variable reftimeVar = nc.findVariable("reftime");
+      Array reftimeValues = reftimeVar.read();
```
tdrwenski: This and the below `Array`s are not used. Is there something reasonable we can assert here? Or did you just want to read them to check no exception was thrown?

tdrwenski (changes requested):
I ran all the tests on Jenkins against your branch and one is failing: https://jenkins-aws.unidata.ucar.edu/view/Users/job/tara-netcdf-java/35/

In the `TestGribCompareBuilders` tests it looks like it is comparing the builder/ non-builder versions for all files in `../grib/src/test/data/` and it looks like there's a difference for the new file you added. I do not believe this is actually related to your changes (I ran the test with that file on `maint-5.x` and it also failed). Since it's unrelated and I'm not sure how important it is that the with/without builders return identical results, maybe it's easiest to just ignore that new test file in that test and possibly fix it later.

grib/src/main/java/ucar/nc2/grib/collection/GribTimeCoordIntervalUtils.java
ethanrd: Yes, I added that thinking there might be situations in which we'd want to handle these different cases. Didn't find any (didn't look too hard) nor did I think too much about how to implement.

grib/src/main/java/ucar/nc2/grib/collection/GribTimeCoordIntervalUtils.java
ethanrd: Good idea.

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
ethanrd: I think I put those in there for debugging and forgot to remove them. I don't think it's worth trying to come up with tests. The strict monotonicity of `timeOffset` is the root issue here.

ethanrd:
Agree, probably easier to ignore the new test file. Are there other test files excluded from this test?

grib/src/test/java/ucar/nc2/grib/coord/TestDiscontiguousInterval.java
ethanrd: Thanks @tdrwenski ! I just pushed a change to remove the unused variable reads.

tdrwenski:
> Agree, probably easier to ignore the new test file. Are there other test files excluded from this test?

Just pushed a fix to ignore that test case!

tdrwenski (approved):

PR #1273, opened 2023-12-13
minor changes for performance

haileyajohnson:
## Description of Changes

_Erase this and add a general description of changes, heads-up on any tricky parts, etc., here._

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
@@ -287,17 +287,17 @@ public class VariableDS extends Variable implements VariableEnhanced, EnhanceSca
 
       double[] dataArray = (double[]) data.get1DJavaArray(DataType.DOUBLE);
 
-      List<Double> list = Arrays.stream(dataArray).parallel().map((num) -> {
+      dataArray = Arrays.stream(dataArray).parallel().map((num) -> {
         for (Enhancement e : toApply) {
           num = e.convert(num);
         }
         return num;
-      }).boxed().collect(Collectors.toList());
+      }).toArray();
 
       Array out = Array.factory(convertedType, data.getShape());
       IndexIterator iterOut = out.getIndexIterator();
-      for (int i = 0; i < list.size(); i++) {
-        iterOut.setObjectNext(list.get(i));
+      for (int i = 0; i < data.getSize(); i++) {
+        iterOut.setObjectNext(dataArray[i]);
       }
       return out;
     }
--- a/cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
@@ -24,6 +24,8 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.DoubleStream;
 
 /**
  * A wrapper around a Variable, creating an "enhanced" Variable. The original Variable is used for the I/O.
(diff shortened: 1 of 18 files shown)


cdm/core/src/main/java/ucar/nc2/filter/ConvertMissing.java
```diff
@@ -189,12 +187,7 @@ public boolean isMissingValue(double val) {
     if (!hasMissingValue) {
       return false;
     }
-    for (double aMissingValue : missingValue) {
-      if (Misc.nearlyEquals(val, aMissingValue, Misc.defaultMaxRelativeDiffFloat)) {
-        return true;
-      }
-    }
-    return false;
+    return Arrays.stream(missingValue).anyMatch(mVal -> Misc.nearlyEquals(val, mVal, Misc.defaultMaxRelativeDiffFloat));
```
tdrwenski: this one could also use `DoubleMath.fuzzyEquals`

haileyajohnson: well I replace the implementation of `Misc.nearlyEquals` with 'DoubleMath.fuzzyEquals`

tdrwenski: oops, missed that!

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -261,23 +258,48 @@ Array convert(Array data, Set<NetcdfDataset.Enhance> enhancements) {
...
+        return num;
+      }).boxed().collect(Collectors.toList());
+
+      Array out = Array.factory(convertedType, data.getShape());
+      IndexIterator iterOut = out.getIndexIterator();
+      for (int i = 0; i < list.size(); i++) {
+        iterOut.setObjectNext(list.get(i));
       }
-      return data;
+      return out;
```
tdrwenski: I don't think you have to make a List and then an Array to return the converted data. This may be what is using more memory than before. I _think_ you can just collect to a primitive array instead of a `List` and use that when creating an `Array`, something like:
```
      double[] out = Arrays.stream(dataArray).parallel().map((num) -> {
        for (Enhancement e : toApply) {
          num = e.convert(num);
        }
        return num;
      }).toArray();

      return Array.factory(convertedType, data.getShape(), out);
```

tdrwenski: on second thought that only works if the `out` type is always double 🤔

tdrwenski: so maybe your solution is good as is except we don't want all three arrays in memory at the same time. So perhaps we can just do `dataArray = null` after creating `list` (or another solution, maybe make it go out of scope somehow)

tdrwenski: It may be better to still create the `double[]` from the stream with `toArray()` like I suggested above since then no boxing needs to happen. And then still iterate over `out` to create the final array of the correct type. Sorry for putting too many comments about this 😆

haileyajohnson: `toArray` only turns the stream back into an array, it doesn't reference the mapped data. You have to `collect` to get the mapped values

cdm/core/src/main/java/ucar/nc2/filter/ConvertMissing.java
```diff
@@ -111,42 +112,62 @@ public static ConvertMissing createFromVariable(VariableDS var) {
             // TODO add logger
           }
         }
-        hasMissingValue = true;
       } else { // not a string
         missingValue = new double[missingValueAtt.getLength()];
         DataType missingType = FilterHelpers.getAttributeDataType(missingValueAtt, signedness);
         for (int i = 0; i < missingValue.length; i++) {
+          double val = var.convertUnsigned(missingValueAtt.getNumericValue(i), missingType).doubleValue();
```
tdrwenski: I don't think this is being used anymore
```suggestion
```

cdm/core/src/main/java/ucar/nc2/filter/ConvertMissing.java
```diff
@@ -111,42 +112,62 @@ public static ConvertMissing createFromVariable(VariableDS var) {
...
     this.hasFillValue = hasFillValue;
     this.fillValue = fillValue;
-    this.hasMissingValue = hasMissingValue;
     this.missingValue = missingValue;
+    this.hasMissingValue = false;
+    // clean up missing values: remove NaNs, fill values, and values outside valid range
+    if (this.missingDataIsMissing && this.missingValue != null) {
+      List<Double> missing = new ArrayList();
+      for (double mv : this.missingValue) {
+        if (mv == Double.NaN) {
```
tdrwenski: I think this should be
```suggestion
        if (Double.isNaN(mv)) {
```

cdm/core/src/main/java/ucar/nc2/dataset/VariableDS.java
```diff
@@ -11,19 +11,16 @@
...
+import ucar.nc2.filter.*;
 import ucar.nc2.internal.dataset.CoordinatesHelper;
 import ucar.nc2.iosp.netcdf3.N3iosp;
 import ucar.nc2.util.CancelTask;
 import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
```
tdrwenski: I don't think these are used anymore
```suggestion
```

tdrwenski (approved):

PR #1275, opened 2023-12-19
Cf anyPoint writer rework to use multiple altNames and timeNames

mnlerman:
## Description of Changes

Addresses the netcdf-java side of https://github.com/Unidata/tds/issues/269

need to build lists of the stations, feaature data and middle data for collections of features. Loop over them when writing to include all information about the collection.

Note possible performance improvement: since we're looping, we're trying to add some variables twice which is being caught and raising a warning that the variable already exists. Could check if the variable exists sooner. 

Still to do: update TDS code to use new methods that have the altName and timeName (I'm working on this today and next week)

## PR Checklist
- [x] Add labels
       until ready for review


cdm/core/src/main/java/ucar/nc2/ft/point/writer/WriterCFStationCollection.java
```diff
@@ -63,14 +61,13 @@ public WriterCFStationCollection(String fileOut, List<Attribute> atts, List<Vari
         "Timeseries of station data in the indexed ragged array representation, H.2.5"));
   }
 
-  public void writeHeader(StationTimeSeriesFeatureCollection stations) throws IOException {
+  public void writeHeader(StationFeatureCollection stations) throws IOException {
```
mnlerman, line 64: This is modifying the signature of a public class but StationTimeSeriesFeatureCollection extends StationFeatureCollection so it shouldn't cause any problems.

haileyajohnson (approved):

Issue #1278, opened 2023-12-24
Lambert conformal projection does not respect units of the false northing/easting


luca009:
### Versions impacted by the bug

5.5.3

### What went wrong?

Trying to use any of the projection functions of a GridCoordSystem (like findXYindexFromLatLon) on a file with lambert confomal projection (or potentially other projections too, I have not tested this however) can result in incorrect coordinates being returned, as the projection functions don't respect the units of the dataset.
In particular, the issue is the hardcoded earth radius, which is given in km by default. However, this does not get converted when the dataset's false northing/easting values are not given in km.
Because of this mismatch of units, the results of the projection functions are always incorrect.

Example datasets where this happens are available from GeoSphere Austria/ZAMG for instance (INCA weather model): [ftp://eaftp.zamg.ac.at/nowcast](ftp://eaftp.zamg.ac.at/nowcast). These files have their false northing/easting given in meters instead of km, which breaks the projection calculations of netcdf-java.

Fix: Check the dataset's units and accordingly convert the earth radius to those units when initializing the GridCoordSystem.

### If you have an example file that you can share, please attach it to this issue.
If so, may we include it in our test datasets to help ensure the bug does not return once fixed?
Note: the test datasets are publicly accessible without restriction.

Yes

### Code of Conduct

- [x] I agree to follow the UCAR/Unidata Code of Conduct

PR #1279, opened 2023-12-25
Fix unit conversion for projection

luca009:
## Description of Changes

Due to an issue with the codebase, the units of the x and y axes cannot be checked until the classes handling projection are already initialized. This results in the issue that the projection conversion always assumes that units are in km, when they might not be (see #1278).
This PR is a fix for this issue, but it definitely needs to be replaced with a better approach some time later. As mentioned, this seems to be a rather deeply rooted issue with the way a NetcdfDataset gets initialized. Judging by some comments, this also [isn't a new problem](https://github.com/Unidata/netcdf-java/blob/f2ad9739f2faa446315f7197f194e966887f228a/cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java#L1616C1-L1619C59).

## PR Checklist

- [ ] Add labels
- [X] Open as a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/)
       until ready for review

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/internal/dataset/CoordinatesHelper.java
+++ b/cdm/core/src/main/java/ucar/nc2/internal/dataset/CoordinatesHelper.java
@@ -58,8 +58,8 @@ public class CoordinatesHelper {
     addAxes(ncd.getRootGroup(), axes);
     coordAxes = ImmutableList.copyOf(axes);
 
-    coordTransforms =
-        builder.coordTransforms.stream().map(ct -> ct.build(ncd, coordAxes)).filter(Objects::nonNull).collect(Collectors.toList());
+    coordTransforms = builder.coordTransforms.stream().map(ct -> ct.build(ncd, coordAxes)).filter(Objects::nonNull)
+        .collect(Collectors.toList());
 
     coordTransforms.addAll(builder.verticalCTBuilders.stream().map(ct -> ct.makeVerticalCT(ncd))
         .filter(Objects::nonNull).collect(Collectors.toList()));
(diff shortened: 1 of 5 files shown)


cdm/core/src/main/java/ucar/nc2/internal/dataset/CoordinatesHelper.java
```diff
@@ -59,7 +59,7 @@ private CoordinatesHelper(Builder builder, NetcdfDataset ncd) {
     coordAxes = ImmutableList.copyOf(axes);
 
     coordTransforms =
-        builder.coordTransforms.stream().map(ct -> ct.build(ncd)).filter(Objects::nonNull).collect(Collectors.toList());
+        builder.coordTransforms.stream().map(ct -> ct.build(ncd, coordAxes)).filter(Objects::nonNull).collect(Collectors.toList());
```
luca009: This is probably the most important line, since we're passing in the already calculated coordinate axes, before the NetcdfDataset has them initialized. Since we need the coordinate axes to build the coordinate transforms correctly, we need to pass them in separately.

haileyajohnson:
Hi @luca009 , thank you so much for your contribution! There are two things this pull request needs to be merged:
1. You've changed the method signature of several public methods/constructors - we can't do that without changing the major version, but we can merge added overload methods
2. To pass the code style check, run the gradle task `spotlessApply`

luca009:
> Hi @luca009 , thank you so much for your contribution! There are two things this pull request needs to be merged:
> 
>     1. You've changed the method signature of several public methods/constructors - we can't do that without changing the major version, but we can merge added overload methods
> 
>     2. To pass the code style check, run the gradle task `spotlessApply`

Should all be fixed up now, thank you for the kind reminders. This is my first time contributing to another project :)

haileyajohnson (approved):

PR #1300, opened 2024-01-31
Improve S3 Zarr performance

tdrwenski:
## Description of Changes

There is no way to control the order of the IOSP classes loaded by the `ServiceLoader`. This can result in very slow performance for S3 Zarr files, since many IOSPs will want to read the data to check if it is a valid file of their type. The `ZarrIosp` check is super fast since it only needs to check whether its Raf is a directory type.

I propose to add a new method to the `IOServiceProvider` interface that will sort the IOSPs into two (or later more) groups so that the fast/common IOSPs can be checked first and the rest checked later.

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/iosp/IOServiceProvider.java
+++ b/cdm/core/src/main/java/ucar/nc2/iosp/IOServiceProvider.java
@@ -35,7 +35,7 @@ import ucar.unidata.io.RandomAccessFile;
  *
  * @author caron
  */
-public interface IOServiceProvider {
+public interface IOServiceProvider extends Comparable<IOServiceProvider> {
 
   /**
    * Check if this is a valid file for this IOServiceProvider.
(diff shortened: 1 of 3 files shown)


cdm/core/src/main/java/ucar/nc2/iosp/IOServiceProvider.java
```diff
@@ -47,6 +47,16 @@ public interface IOServiceProvider {
    */
   boolean isValidFile(RandomAccessFile raf) throws IOException;
 
+  /**
+   * Check if this IOSP's isValidFile(raf) is a fast operation.
+   * If so, it will be checked earlier than the slower dynamically loaded IOSPs.
+   * For instance, file reads make isValidFile(raf) a slower check, especially for S3 files.
+   * @return true if isValidFile(raf) is a fast operation, else false.
+   */
+  default boolean isValidFileIsFast() {
```
tdrwenski: not sure about this name... instead of breaking the IOSPs into fast/slow groups, we could choose to break into "ones to check earlier" (regardless of speed) and "ones to check later". Then maybe call this `checkFirst`/`checkEarlier` or something

haileyajohnson (approved):
I like it :)

PR #1308, opened 2024-02-26
Update dts war version used in testing

tdrwenski:
## Description of Changes

I'm not sure why we test opendap this way, but we use gretty and the TDS's dts.war to test the opendap client. This version was badly out of date, so this PR updates it. Tests are passing on Jenkins on this branch: https://jenkins-aws.unidata.ucar.edu/view/Users/job/tara-netcdf-java/49/

### DIFF
--- a/gradle/any/gretty.gradle
+++ b/gradle/any/gretty.gradle
@@ -72,10 +72,10 @@ gretty {
 farm {
   jvmArgs = ["-Dlog4j.configurationFile=$rootDir/gradle/gretty/log4j2Config.xml"]
   // used by :dap4 test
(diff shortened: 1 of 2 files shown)


tdrwenski:
TDS 5.4 uses JDK 11 and so cannot be used by netcdf-java and JDK 8

DennisHeimbigner:
This way of testing is for historical reasons. Ideally it should be changed to match
the way DAP4 uses remotetest.unidata.ucar.edu instead of gretty (assuming that latter approach
is a better way to test :-)

tdrwenski:
Thanks for the comment @DennisHeimbigner! Was just re-investigating why Netcdf-java needs gretty and came across this.

PR #1315, opened 2024-03-12
revert switch to fuzzy equals

### LINKED ISSUE #1312: Unpacked values from scaled/offset dataset are all NaN.

Any build of version 5.5.4 from the past month.

### What went wrong?

A user wrote that a plot of his data was coming up blank when using a just-downloaded Panoply. A quick check revealed that Panoply was seeing all NaN values even though an ncdump showed good values. Some further testing demonstrated that this was also occurring with a copy of Panoply from a month ago, but that a copy from the very end of December 2023 was plotting as expected. In all three cases, the Panoply release was using an up-to-date build of the netcdf-Java 5.x-maint snapshot.

The dataset packs the data as shorts and specifies a scale value and an offset. Panoply acquires the dataset in _enhanced_ mode and accepts whatever values netCDF-Java says are there.

If I modify Panoply to acquire the dataset in NOT enhanced mode, then netCDF-Java reports the unpacked short values and a plot is successfully made. If I build a copy of Panoply using a copy of netcdfAll that I built on December 28 and tell it to acquire the dataset in enhanced mode, the unpacked data values are correctly reported and a plot is successfully made.

Looking at the commit history on the 5.x-maint branch, it looks like there were some modifications to unpacking data made about January 12, which seems a likely point where the problem is occurring. Nothing I noticed in the changes then jumps out at me as the obvious cause, but I haven't dug very deep yet.

Not sure if it's at all meaningful, but the scale and offset values are reeeaally small.

A copy of the example dataset that revealed the issue can be downloaded from [Dropbox](https://www.dropbox.com/scl/fi/52hie61ph6rmvn0xy8x4f/eccc2016.nc?rlkey=ad35agfz6srl9feod12b064hb&dl=1). It's a CF-1.6 dataset with lon-lat data on a relatively small array covering the re4gion around Mexico City.

### Relevant stack trace

_No response_

### Relevant log messages

_No response_

Yes

### Code of Conduct

haileyajohnson:
Fixes #1312

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/util/Misc.java
+++ b/cdm/core/src/main/java/ucar/nc2/util/Misc.java
@@ -97,8 +97,8 @@ public class Misc {
   }
 
   /** AbsoluteDifference is less than maxAbsDiff. */
-  public static boolean nearlyEquals(float a, float b, float maxAbsDiff) {
-    return DoubleMath.fuzzyEquals(a, b, maxAbsDiff);
+  public static boolean nearlyEquals(float a, float b, float maxRelDiff) {
+    return relativeDifference(a, b) < maxRelDiff;
   }
 
   /** AbsoluteDifference is less than {@link #defaultMaxRelativeDiffDouble}. */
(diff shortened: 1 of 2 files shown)


cdm/core/src/main/java/ucar/nc2/util/Misc.java
```diff
@@ -97,8 +97,8 @@ public static boolean nearlyEquals(float a, float b) {
   }
 
   /** AbsoluteDifference is less than maxAbsDiff. */
```
tdrwenski:
```suggestion
  /** RelativeDifference is less than maxRelDiff. */
```

cdm/core/src/main/java/ucar/nc2/util/Misc.java
```diff
@@ -107,8 +107,8 @@ public static boolean nearlyEquals(double a, double b) {
...
```
tdrwenski:
```suggestion
  /** RelativeDifference is less than maxRelDiff. */
```

tdrwenski (approved):

PR #1318, opened 2024-03-15
Convert missing values before scale and offset

haileyajohnson:
Swap the order that `ConvertMissing` and `ScaleOffset` are applied so that `ConvertMissing` uses packed values to check for missing values.

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/filter/ConvertMissing.java
+++ b/cdm/core/src/main/java/ucar/nc2/filter/ConvertMissing.java
@@ -61,27 +61,15 @@ public class ConvertMissing implements Enhancement {
       }
     }
 
-    /// fill_value
-    boolean hasFillValue = var.hasFillValue();
-    double fillValue = var.getFillValue();
-    // need fill value info before convertMissing
-    Attribute fillValueAtt = var.findAttribute(CDM.FILL_VALUE);
-    if (fillValueAtt != null && !fillValueAtt.isString()) {
-      DataType fillType = FilterHelpers.getAttributeDataType(fillValueAtt, var.getSignedness());
-      fillValue = var.convertUnsigned(fillValueAtt.getNumericValue(), fillType).doubleValue();
-      hasFillValue = true;
-    } else {
-      // No _FillValue attribute found. Instead, if file is NetCDF and variable is numeric, use the default fill value.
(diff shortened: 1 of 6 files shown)


cdm/core/src/main/java/ucar/nc2/filter/ConvertMissing.java
```diff
@@ -208,7 +185,7 @@ public double getFillValue() {
 
   public boolean isMissingValue(double val) {
     for (double aMissingValue : missingValue) {
-      if (Misc.nearlyEquals(val, aMissingValue, Misc.defaultMaxRelativeDiffFloat)) {
+      if (val == aMissingValue) {
```
tdrwenski, line 188: I'm not completely sure about the use of `==` here and in `isFillValue`:
- Just to be sure, `aMissingValue` or `fillValue` are not be allowed to be `NaN` (or in the case of `fillValue`, then `hasFillValue` will be `false` at least), right?
-  But also, are we sure that exact matching for doubles is okay? If any operations manipulate these doubles (say they were packed and then unpacked) they may not match exactly

cdm/core/src/test/java/ucar/nc2/dataset/TestScaleOffsetMissingUnsigned.java
```diff
@@ -286,7 +286,8 @@ public void testScaleOffsetMissingUnsigned() throws URISyntaxException, IOExcept
 
   // This test demonstrated the bug in https://github.com/Unidata/netcdf-java/issues/572, but for unsigned variables.
   @Test
-  public void testNegativeScaleOffsetValidRangeUnsigned() throws URISyntaxException, IOException {
+  public void testNegativeScaleOffsetValidRangeUnsigned()
+      throws URISyntaxException, IOException, InvalidRangeException {
```
tdrwenski: Don't think this change is still needed
```suggestion
  public void testNegativeScaleOffsetValidRangeUnsigned() throws URISyntaxException, IOException {
```

cdm/core/src/main/java/ucar/nc2/filter/ConvertMissing.java
tdrwenski: Discussed via slack

tdrwenski (approved):

PR #1320, opened 2024-03-22
Don't open all zips as zarr

### LINKED ISSUE #1319: ZarrIosp claims any zipped dataset
From issue: https://github.com/Unidata/netcdf-java/issues/1307:

When using netcdfAll with the optional cdm-zarr code included, any zipped dataset you try to open gets claimed by `ZarrIosp` as a validFile. The zip is apparently opened, but if it's not actually zarr, then it is reported as empty.

tdrwenski:
## Description of Changes

Fixes https://github.com/Unidata/netcdf-java/issues/1319.

Currently, the `ZarrIosp` opens all files that have the Raf type `RandomAccessDirectory`. The Raf is assigned this type for in the case `MFiles.create(location).isDirectory() == true`.

For a single zipped file, we would like Zarr to not be used. So we need the Raf to not be a `RandomAccessDirectory`. Then, as without the zarr code, the compressed file will get uncompressed and the uncompressed file will be assigned a normal `RandomAccessFile` Raf.

This PR fixes this issue by not classifying single zipped files as "directories".


tdrwenski:
Jenkins tests passing: https://jenkins-aws.unidata.ucar.edu/view/Users/job/tara-netcdf-java/52/

haileyajohnson (approved):

PR #1329, opened 2024-04-10
Use TestContainers

tdrwenski:
## Description of Changes

The goal of this PR is to replace usage of remotetest and gretty in netcdf-java with TestContainers. The TestContainer uses docker to start/stop the dts and d4ts war files.

Additional notes:

- Originally I tried to use a category for all tests that now need docker to be able to choose to exclude them. However, this did not work correctly with `@Parameterized` tests, as these classes are still initialized even when the category would make them excluded. If they refer to the docker server in their static fields then they would fail to initialize due to docker not being present. So here I use the less nice solution of excluding the entire package of tests that need docker when `-DskipDockerTests=True` is given.
- In the docker file, when I used the d4ts-5.4.war file there were some issues which were resolved with the 5.5-SNAPSHOT. That snapshot, however, will only be on nexus for a short while so we will have to update this in the next few weeks. Hopefully soon we can switch this to the 5.5 release.
- For now we will have to exclude the docker tests on Jenkins using `-DskipDockerTests=True`


haileyajohnson (approved):
🎉

PR #1333, opened 2024-04-15
Fix renaming a dimension in aggregations

tdrwenski:
## Description of Changes

Renaming an aggregation dimension using NcML does not work-- it will throw an exception in the case of an aggregation with a scan element (see https://github.com/Unidata/tds/issues/490) and have the wrong size in the case of an aggregation with locations. The aggregation dimension is already part of the `groupBuilder`, so we need to use that instead of the `refGroup` (which is the dimension from a single file, or null in the case of a scan), in order to treat it as an existing dimension instead of a new one.

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/internal/ncml/NcmlReader.java
+++ b/cdm/core/src/main/java/ucar/nc2/internal/ncml/NcmlReader.java
@@ -810,9 +810,11 @@ public class NcmlReader {
 
     String nameInFile = dimElem.getAttributeValue("orgName") != null ? dimElem.getAttributeValue("orgName") : name;
 
-    // LOOK this is wrong, groupBuilder may already have the dimension.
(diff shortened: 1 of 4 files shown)


cdm/core/src/main/java/ucar/nc2/internal/ncml/NcmlReader.java
```diff
@@ -810,9 +810,11 @@ private void readDim(Group.Builder groupBuilder, @Nullable Group refGroup, Eleme
 
     String nameInFile = dimElem.getAttributeValue("orgName") != null ? dimElem.getAttributeValue("orgName") : name;
 
-    // LOOK this is wrong, groupBuilder may already have the dimension.
```
haileyajohnson, line 813: 🤣

haileyajohnson (approved):

PR #1335, opened 2024-04-19
GDAL_NoData geotiff tag should be a string, not float

WeatherGod:
* See https://gdal.org/drivers/raster/gtiff.html#nodata-value

## Description of Changes

When greyscale is `false`, the data is scaled to the range of 1-255 before being written to the geotiff. The encoding value of 0 is reserved for missing/nodata values. The missing value is encoded in the GDAL_NoData geotiff tag, but we've been encoding it incorrectly all this time. This PR corrects that, as well as expands the geotiff tests to exercise both cases of greyscale being True and False.

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
+++ b/cdm/misc/src/main/java/ucar/nc2/geotiff/GeotiffWriter.java
@@ -223,7 +223,7 @@ public class GeotiffWriter implements Closeable {
       float max = (float) (dataMinMax.max);
       geotiff.addTag(new IFDEntry(Tag.SMinSampleValue, FieldType.FLOAT).setValue(min));
       geotiff.addTag(new IFDEntry(Tag.SMaxSampleValue, FieldType.FLOAT).setValue(max));
-      geotiff.addTag(new IFDEntry(Tag.GDALNoData, FieldType.FLOAT).setValue(min - 1.f));
+      geotiff.addTag(new IFDEntry(Tag.GDALNoData, FieldType.ASCII).setValue(String.valueOf(min - 1.f)));
     }
 
     /*
--- a/cdm/misc/src/test/java/ucar/nc2/geotiff/TestGeoTiffWriter.java
+++ b/cdm/misc/src/test/java/ucar/nc2/geotiff/TestGeoTiffWriter.java
@@ -16,6 +16,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import ucar.ma2.Array;
 import ucar.ma2.InvalidRangeException;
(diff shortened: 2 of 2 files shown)


WeatherGod:
Failing test is unrelated as it pertains to zarr, and it might be transient?

ethanrd:
Hi @WeatherGod - What versions of the GeoTiff Specification ([1.0](http://geotiff.maptools.org/spec/geotiffhome.html)* or [1.1](https://docs.ogc.org/is/19-008r4/19-008r4.html)) are you targeting? (I'm not certain what netCDF-java currently supports but suspect it originally targeted 1.0.) It looks like GDAL supports both GeoTIFF 1.0 and 1.1, defaulting to 1.0 (see GEOTIFF_VERSION in GDAL docs [Configuration Options section](https://gdal.org/drivers/raster/gtiff.html#configuration-options)).

I'm not familiar with the differences. From the GDAL docs, maybe just "clear[ing up] ambiguities and fix[ing] inconsistencies". Either way, it's probably best to reference the specification documents than the GDAL documentation to ensure broad interoperability.

[*] FYI There's an illegible TOC button/image at the top of the v1.0 page that, embarrassingly, took me awhile to find.

WeatherGod:
Actually, neither. GDAL_NoData is an extended TIFF tag that has been widely adopted. So much so that it is recognized by the Library of Congress: https://www.loc.gov/preservation/digital/formats/content/tiff_tags.shtml. It says it is ASCII encoded.

WeatherGod:
By the way, I've also come across another set of bugs related to the encoding of the IFDEntry tags. It looks like it doesn't handle unsigned bytes correctly. None of that is tested (until my other PR exposed it).

WeatherGod:
As for the IFDEntry bug, I can't exercise it explicitly with the current API, so I'm just going to have to punt that and have it get fixed by my other PR that exercises it by saving unsigned bytes data. Let me know if you need anything else for this PR.

ethanrd:
Sorry, I was trying to ask a more general question beyond the GDAL_NoData attribute/tag but didn't find a more general issue to comment on.

Support for "official" standards is a high-level goal for netCDF-Java (and Unidata more generally). Of course, full support for standards isn't always possible and community/implementation conventions are often needed (in addition to standards or as the main focus) for the broadest level of interperability (or simply because that's what gets implemented).

My question was, I guess, mainly about what specification you are using as a reference in your work on the netCDF-Java GeoTIFF code. It sounds like you are using GDAL and LoC GeoTIFF documentation as your main specification. So, I think I have my answer.

Thanks for all your work and contributions!

tdrwenski (approved):

PR #1341, opened 2024-05-06
Fix and extend Geotiff IFD handling to allow all integer types

WeatherGod:
## Description of Changes
* Follows specs from pg. 15-16 of v6.0 TIFF: https://www.itu.int/itudoc/itu-t/com16/tiff-fx/docs/tiff6.pdf
* Added roundtrip tests of the private Geotiff method via reflection
* This does reveal an existing API flaw that will cause overflow of large unsigned integers

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/misc/src/test/java/ucar/nc2/geotiff/TestIFDEntry.java
+++ b/cdm/misc/src/test/java/ucar/nc2/geotiff/TestIFDEntry.java
@@ -11,9 +11,7 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.lang.ReflectiveOperationException;
 import java.lang.invoke.MethodHandles;
-import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
@@ -79,22 +77,17 @@ public class TestIFDEntry {
   }
 
   @Test
-  public void testRoundtrip() throws ReflectiveOperationException {
-    GeoTiff geotiff = new GeoTiff("foobar");
+  public void testRoundtrip() {
     // 16 bytes should be more than enough
     ByteBuffer buffer = ByteBuffer.allocate(16);
     ByteOrder byteOrder = ByteOrder.BIG_ENDIAN;
     buffer.order(byteOrder);
 
-    Method writeMethod =
-        geotiff.getClass().getDeclaredMethod("writeIntValue", ByteBuffer.class, ifd.getClass(), int.class);
-    Method readMethod = geotiff.getClass().getDeclaredMethod("readIntValue", ByteBuffer.class, ifd.getClass());
-    readMethod.setAccessible(true);
-    writeMethod.setAccessible(true);
-    int writeSize = (int) writeMethod.invoke(geotiff, buffer, ifd, testValue);
+    int writeSize = GeoTiff.writeIntValue(buffer, ifd, testValue);
+    Assert.assertEquals(ifd.type.size, writeSize);
     buffer.position(0);
 
-    int readValue = (int) readMethod.invoke(geotiff, buffer, ifd);
+    int readValue = GeoTiff.readIntValue(buffer, ifd);
 
     Assert.assertEquals(testValue, readValue);
   }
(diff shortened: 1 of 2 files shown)


WeatherGod:
Was talking with a colleague about how I couldn't test private methods, which was needed to properly test out the fixes to read/writeIntValue, when they pointed me to Java reflection, which I never knew about.

cdm/misc/src/test/java/ucar/nc2/geotiff/TestIFDEntry.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Test
+  public void testRoundtrip() throws ReflectiveOperationException {
+    GeoTiff geotiff = new GeoTiff("foobar");
+    // 16 bytes should be more than enough
+    ByteBuffer buffer = ByteBuffer.allocate(16);
+    ByteOrder byteOrder = ByteOrder.BIG_ENDIAN;
+    buffer.order(byteOrder);
+
+    Method writeMethod =
```
tdrwenski: I personally don't like this reflection solution for testing private methods. First, in general, private methods should not be tested, though I understand that in practice some classes would need big refactors to make the public API unit testable. Second, I think it makes the test harder to read. And finally, if someone were to rename `writeIntValue`, the compiler would not catch an issue here, you would just get `NoSuchMethodException`.

Is there any reasonable way to test the public API of `GeoTiff` instead? I am not as familiar with the code, so I am not sure how feasible this is. If not, I propose to make the method `writeIntValue` package private so you can test it without reflection. Also not a perfect solution, but I think it is slightly nicer.

cdm/misc/src/test/java/ucar/nc2/geotiff/TestIFDEntry.java
```diff
@@ -0,0 +1,104 @@
...
+
+package ucar.nc2.geotiff;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.io.*;
```
tdrwenski: Unused import:
```suggestion
```

cdm/misc/src/test/java/ucar/nc2/geotiff/TestIFDEntry.java
```diff
@@ -0,0 +1,104 @@
...
+import java.lang.ReflectiveOperationException;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.ArrayList;
+import java.util.Collections;
```
tdrwenski: Unused import:
```suggestion
```

cdm/misc/src/test/java/ucar/nc2/geotiff/TestIFDEntry.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @author Ben Root
+ * @since 5/6/2024
+ */
+@RunWith(Parameterized.class)
+public class TestIFDEntry {
+  private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  public IFDEntry ifd;
+  // Because we are dealing with reflection
+  public int testValue;
```
tdrwenski: I think these can be `private final`
```suggestion
  private final IFDEntry ifd;
  private final int testValue;
```

tdrwenski (changes requested):
Thanks for this PR @WeatherGod! I just had a comment about the use of reflection and some very minor style issues in the tests.

cdm/misc/src/test/java/ucar/nc2/geotiff/TestIFDEntry.java
WeatherGod: Coming from python, the idea that one shouldn't test private methods is completely foreign to me. Unfortunately, with the way the GeoTiff class is designed, it would be difficult to not only construct the many geotiffs instances with the appropriate data types and ranges, but to then subsequently test that it was encoded correctly. The test would have a huge amount of unrelated setup that would make the test difficult to read and understand.

I am totally fine with changing it to be package private. We could even go a step further and make them static as they do not interact with any class members.

tdrwenski (approved):
Looks good! Thanks!

PR #1342, opened 2024-05-09
Fix a bug with geotiff writing when yaxis is flipped

WeatherGod:
## Description of Changes

Correctly compute yStart when flipping the y-axis. Not doing can result in the bounding box to be in the wrong location if the source coordinates were in kilometers.

## PR Checklist

- [ ] Add labels
       until ready for review


WeatherGod:
... I swear, one of these days I'll remember to use spotlessApply before committing...

tdrwenski (approved):
Thanks @WeatherGod!

WeatherGod:
Just to add a clarifying comment here (and I'll correct the comment in the code in my other PR), I got my sense of "flipped" flipped around. The test data is special because its y-axis goes from negative to positive, which is the condition in which the geotiff writer will flip the data around.

PR #1358, opened 2024-06-27
Suppress CVEs

tdrwenski:
## Description of Changes

Suppress three spurious and not applicable jfreechart CVEs in ToolsUI component.

### DIFF
--- a/project-files/owasp-dependency-check/dependency-check-suppression.xml
+++ b/project-files/owasp-dependency-check/dependency-check-suppression.xml
@@ -24,4 +24,14 @@
     ]]></notes>
     <cve>CVE-2023-35116</cve>
   </suppress>
+  <suppress>
+    <notes><![CDATA[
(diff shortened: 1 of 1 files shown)


project-files/owasp-dependency-check/dependency-check-suppression.xml
```diff
@@ -24,4 +24,28 @@
     ]]></notes>
     <cve>CVE-2023-35116</cve>
   </suppress>
+  <suppress>
+    <notes><![CDATA[
```
haileyajohnson, line 28: I think it works to have all theses grouped in a single <suppress> with multiple <vulnerbilityName> elements

tdrwenski: okay cool, let me try that

tdrwenski: done!

haileyajohnson (approved):

PR #1360, opened 2024-07-09
revert axis name changes and add work done to use coordVars to set co…

haileyajohnson:
- reverts time and alt axis names changes from #1275 to fix DSGSubsetWriter for NCSS
- adds constructors that contain a list of coordinate variables to point feature implementations and point writers
- future work to make the CF Point writers more accurately represent the dataset they're writing should use coordinate axis info taken from the list of coordinate variables

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/ft/point/ProfileFeatureImpl.java
+++ b/cdm/core/src/main/java/ucar/nc2/ft/point/ProfileFeatureImpl.java
@@ -6,9 +6,12 @@ package ucar.nc2.ft.point;
 
 import javax.annotation.Nonnull;
 import ucar.nc2.constants.FeatureType;
+import ucar.nc2.dataset.CoordinateAxis;
 import ucar.nc2.ft.ProfileFeature;
 import ucar.nc2.time.CalendarDateUnit;
 import ucar.unidata.geoloc.LatLonPoint;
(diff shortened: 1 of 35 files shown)


cdm/core/src/main/java/ucar/nc2/ft/point/ProfileFeatureImpl.java
```diff
@@ -31,9 +34,9 @@ public ProfileFeatureImpl(String name, CalendarDateUnit timeUnit, String altUnit
     }
   }
 
-  public ProfileFeatureImpl(String name, String timeName, CalendarDateUnit timeUnit, String altName, String altUnits,
-      double lat, double lon, double time, int nfeatures) {
-    super(name, timeName, timeUnit, altName, altUnits);
+  public ProfileFeatureImpl(String name, List<CoordinateAxis> coords, double lat, double lon, double time,
```
tdrwenski, line 37: Can we make this `protected` instead of `public` maybe? If so, same question for `StationProfileCollectionImpl`, `StationProfileFeatureImpl`, and others with public constructors. Though they already were `public`, so we don't necessarily need to fix that in this PR, but I am just thinking after the next release that would technically be a breaking change.

tdrwenski (approved):

PR #1373, opened 2024-07-23
Improve performance of enhancement providers

tdrwenski:
## Description of Changes

Improve performance of enhancement providers by only loading once


matakleo (approved):
Looks pretty good to me!

PR #1393, opened 2024-10-14
refactor some of the enhancement SPI lgoic and provider interface

haileyajohnson:
changed and cleaned up the logic of the EnhancementProvider a bit
- each provider now needs to know it's own attribute name
- the logic of whether to apply an enhancement is made by VariableDS
This is meant to partition things such that it's a bit safer to execute runtime-loaded Enhancements.

I'm also looking into making a Vectorize Enhancement, as mentioned in this [issue](https://github.com/Unidata/tds/issues/527), since that's something we want at CDIP as well. If/when I get it working I'll add it as an example in the docs, similar to the lighting IOSP.

### DIFF
--- a/cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
+++ b/cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
@@ -130,11 +130,17 @@ public class NetcdfDataset extends ucar.nc2.NetcdfFile {
      * x<0 --> 0 x>0 --> 1
      */
     ApplyClassifier,
+    /**
+     * All other Enhancement implementations that are loaded by service provider
+     * This includes third party implementations found on the classpath and loaded at runtime.
+     */
+    ApplyRuntimeLoadedEnhancements,
   }
 
-  private static Set<Enhance> EnhanceAll = Collections.unmodifiableSet(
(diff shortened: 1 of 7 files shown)


cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
```diff
@@ -130,11 +130,17 @@ public enum Enhance {
      * x<0 --> 0 x>0 --> 1
      */
     ApplyClassifier,
```
tdrwenski: should we remove `ApplyClassifier`, `ApplyNormalizer`, `ApplyStandardizer` and instead use `ApplyRuntimeLoadedEnhancements` for those classes since those are being runtime loaded?

tdrwenski (approved):
This looks great, thanks! Definitely makes more sense to move the enhancement name to each provider.

cdm/core/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
haileyajohnson: Yeah I kinda think so too, it seems like if anyone has those attributes in their catalogs they specifically would want them to be on, can't see a reason to leave that fine-level control in.

PR #1394, opened 2024-11-08
BUG: Fix decompressing all zlib data for NIDS

dopplershift:
## Description of Changes
Code was only expanding the buffer by 10000 bytes as it read through. This resulted in blowing the array bounds when the last, noncompressed chunk of data was larger than this. This was easily broken by products with a single compressed chunk, followed by all uncompressed data. A representative data file and smoke test is included.

This was actually triggered by a failure to build Siphon docs trying to access this data over CDMRemote.

## PR Checklist

- [ ] Add labels
       until ready for review


lesserwhirls (approved):
Overall looks good! We'll want to see how this works out on the AWS radar server focused TDS to keep an eye on memory usage, but I think we'll be ok.

dopplershift:
FYI, this particular compression should only impact NIDS/Level 3, there should be no impact on the Level 2 decompression.

PR #1419, opened 2025-02-24
Upgrade github tests to thredds-test-action v3

lesserwhirls:
## Description of Changes

Use `Unidata/thredds-test-action` v3 for running the test matrix on GitHub.


lesserwhirls:
First set of tests failed because JDK 21 was not in the test environment. Second set failed because JDK 21 isn't supported by the version of gradle being used in netCDF-Java (expected). Will now remove JDK 21 from the test matrix.

PR #1450, opened 2025-05-08
v5.8.0 Release

lesserwhirls:
Release `v5.8.0` and begin work on `v5.9.0`


rschmunk:
@lesserwhirls It took me way too long to figure out that sometime in the past month, there's been a change in the build process such that the `netcdfAll` jar now lands in `build/distributions` rather than `build/libs`.

lesserwhirls:
Yes, the most recent upgrade to gradle required an upgrade to the shadow plugin, which changed where the uber jars are generated. The [docs were updated](https://docs.unidata.ucar.edu/netcdf-java/current/userguide/building_from_source.html), but I should edit the release notes to highlight it there as well.

Issue #1375, opened 2024-08-09
Units for false_easting and false_northing get converted to km


iteubner:
### Versions impacted by the bug

We are using the latest release of thredds-docker (version 5.5), based on this [commit](https://github.com/Unidata/tds/commit/3c09bf0b12b24a9e8333ed088f8f83d90788cd48), I assume it is version 5.6.0 or 5.6.1-SNAPSHOT. I was not able to identify the version in the container.

### What went wrong?

Using thredds-docker, when subsetting a netcdf that has non-zero false_easting and false_norting and units m (https://epsg.io/3416), e.g. [NetcdfSubset](https://thredds.hub.geosphere.at/thredds/ncss/grid/fuse-at/snowgrid-cl/HNW_SNOWGRID-CL_SDM_CNRM-CERFACS_CNRM-CM5_r1i1p1_rcp45_CLMcom_CCLM4-8-17.nc/dataset.html) in [catalog](https://thredds.hub.geosphere.at/thredds/catalog/s3catalogs/fuse-at_snowgrid-cl.html?dataset=fuse-at/snowgrid-cl/HNW_SNOWGRID-CL_SDM_CNRM-CERFACS_CNRM-CM5_r1i1p1_rcp45_CLMcom_CCLM4-8-17.nc), the units for false_easting and false_northing get converted to km. But the projection_{x|y}_coordinates (and specifications in the esri_pe_string for the data variables) are still in m. 

Definition of projection and projection coordinates in the original file:
```
char lambert_conformal_conic;
  :semi_major_axis = 6378137.0; // double
  :latitude_of_projection_origin = 47.5; // double
  :inverse_flattening = 298.257222101; // double
  :longitude_of_central_meridian = 13.3333333333333; // double
  :standard_parallel = 49.0, 46.0; // double
  :false_northing = 400000.0; // double
  :grid_mapping_name = "lambert_conformal_conic";
  :semi_minor_axis = 6356752.31414036; // double
  :false_easting = 400000.0; // double

double x(x=585);
  :units = "m";
  :long_name = "x coordinate of projection";
  :standard_name = "projection_x_coordinate";
  :axis = "X";

double y(y=314);
  :long_name = "y coordinate of projection";
  :standard_name = "projection_y_coordinate";
  :axis = "Y";
  :units = "m";
```

Values after subsetting (with projection coordinates added again for completeness):
```
int lambert_conformal_conic;
  :grid_mapping_name = "lambert_conformal_conic";
[...]

PR #1455, opened 2025-06-27
#1375 Units for false_easting and false_northing (fix)

adriaanse-deltares:
## Description of Changes

See the description of the issue #1375

I have worked with GIS systems and Geographic projections at Deltares since 1998 and I am 100% sure that:
No exceptions exist where false easting/northing units differ from the projected coordinates. These values are intrinsic to the coordinate system's design and always share the same units

So I propose to remove the scaling of these parameters completely, also for other projections where it is applied.

## PR Checklist

- [ ] Add labels
       until ready for review


lesserwhirls:
I appreciate you taking the time to make this PR, and it certainly gave me some hints when trying to untangle what was happening. As I said over in Unidata/netCDF-Java#1375, we probably need to support the scaling in the read path. However, I was able to get the scaling addressed in the write path when writing from a Coverage, which will fix the issue of the NetCDF Subset Service producing netCDF files with mismatched cartesian coordinate units and false_easting/false_northing values.

PR #1472, opened 2025-08-08
Release 5.9.0

lesserwhirls:
Release v5.9.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,9 +29,9 @@ allprojects {
   // We try to follow semantic versioning, and thus we use <major>.<minor>.<patch>-<prerelease version>
   // <prerelease version> may be SNAPSHOT, alphax, betax, etc.
   // Note - if bumping to a new major or minor version, be sure to update the docs
-  version = '5.9.0'
-  //status = 'development'
-  status= 'release'
+  version = '5.9.1-SNAPSHOT'
+  status = 'development'
+  //status= 'release'
 }
 
 // Matches Maven's "project.description".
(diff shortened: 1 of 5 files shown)


rschmunk:
I noticed a few days ago with a snapshot and also see this now when building 5.9.0 from scratch that the UDUNITS section is showing some warnings. Perhaps due to #1461?

`
> Task :udunits:compileJavacc
Java Compiler Compiler Version 7.0.13 (Parser Generator)
(type "javacc" with no arguments for help)
Reading from file /some/path/netcdf-java-5.9.0/udunits/src/main/javacc/StandardUnitFormat.jj . . .
Warning: Choice conflict in (...)* construct at line 562, column 5.
         Expansion nested within construct and expansion following construct
         have common prefixes, one of which is: <SP>
         Consider using a lookahead of 2 or more for nested expansion.
Warning: Choice conflict in [...] construct at line 570, column 13.
         Expansion nested within construct and expansion following construct
         have common prefixes, one of which is: "."
         Consider using a lookahead of 2 or more for nested expansion.
Warning: Choice conflict in [...] construct at line 594, column 5.
         Expansion nested within construct and expansion following construct
         have common prefixes, one of which is: "+"
         Consider using a lookahead of 2 or more for nested expansion.
`

lesserwhirls:
That's exactly it. I'm almost willing to bet those warning where there the last time these files were generated and committed to the project (in 2009). Now that we actually see them, we have a chance to address them.

PR #1495, opened 2025-10-10
Delete the print hello when accessing grib2 Type42 data

laurentba:
This PR deletes the print hello when accessing grib2 Type42 data


lesserwhirls:
Oops! That one slipped through when I added support for Type42. Thanks for catching that!

lesserwhirls (approved):

Issue #1382, opened 2024-08-26
ucar.nc2.filter.UnknownFilterException: Unknown filter - v5.6.0 


jacobwentzib:
### Versions impacted by the bug

5.6.0 - Latest

**Environment:**
**UCAR Library Version:** cdm-core 5.6.0
**Java Version:** 21 

### What went wrong?

Recently upgraded to 5.6.0 and see the following exception being thrown where the ServiceLoader is unable to load Filters properly. 

https://github.com/Unidata/netcdf-java/blob/42186c77b9d424aeb45e333be80593c7a03c608e/cdm/core/src/main/java/ucar/nc2/filter/Filters.java#L80

Once we downgrade back to _5.5.3_ everything works properly. 

Has there been any changes to classpath or anything else that could be causes this issue? We are seeing this isolated to jar executions but IDE development with target classes runs properly.

We have created a custom ServiceLoader within our application and can see the resources being found. 

### Relevant stack trace

Being invoked from: 
Variable.class:
```java
    public Array read(Section section) throws IOException, InvalidRangeException {
        return section == null ? this._read() : this._read(Section.fill(section, this.shape));
    }
```

Error throw from above link:
_java.io.IOException: ucar.nc2.filter.UnknownFilterException: Unknown filter: no filter found with id 2_

### If you have an example file that you can share, please attach it to this issue.
If so, may we include it in our test datasets to help ensure the bug does not return once fixed?
Note: the test datasets are publicly accessible without restriction.

Yes

### Code of Conduct

- [X] I agree to follow the UCAR/Unidata Code of Conduct

tdrwenski:
- Can you add the full error stack trace?
- Are any FilterProviders getting loaded (yours or NetCDF-java's)? It looks like the error is about "id 2" which should be the `Shuffle` filter.

jacobwentzib:
NetCDF-java's FilterProviders. 

Here is the full trace: 
java.io.IOException: ucar.nc2.filter.UnknownFilterException: Unknown filter: no filter found with id 2
        at ucar.nc2.internal.iosp.hdf5.H5tiledLayoutBB.<init>(H5tiledLayoutBB.java:90) ~[cdm-core-5.6.0.jar!/:5.6.0]
        at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:226) ~[cdm-core-5.6.0.jar!/:5.6.0]
[...]

PR #1555, opened 2026-05-12
Support customised service loader

nicobrevin:
## Description of Changes

As per #1382, adds a custom ServiceLoader#load method that supports customizing the classloader.  In our case we have something of a plugin architecture that isolates the different plugins in their own class loader.  This causes chaos and hilarity when the netcdf plugin is used,  as it is not going to execute that code with the netcdf library's classloader set as the calling thread's context classloader.  I have worked around it to some extent by finaggling our code to do a best-effort and forcibly load some of the offending netcdf classes with the context class loader set correctly, but that only works for the code that loads services in a static initializer.

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/cdm/core/src/main/java/thredds/inventory/MFiles.java
+++ b/cdm/core/src/main/java/thredds/inventory/MFiles.java
@@ -6,7 +6,6 @@
 package thredds.inventory;
 
 import java.io.IOException;
-import java.util.ServiceLoader;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
@@ -14,6 +13,9 @@ import org.slf4j.LoggerFactory;
 import thredds.filesystem.MFileOS;
 import thredds.filesystem.MFileOS7;
 import ucar.nc2.internal.ncml.NcmlReader;
+import ucar.nc2.util.NcServiceLoader;
+
+
 
 /**
  * Static helper methods for MFile objects.
@@ -34,7 +36,7 @@ public class MFiles {
     MFileProvider mFileProvider = null;
 
     // look for dynamically loaded MFileProviders
-    for (MFileProvider provider : ServiceLoader.load(MFileProvider.class)) {
+    for (MFileProvider provider : NcServiceLoader.load(MFileProvider.class)) {
       if (provider.canProvide(location)) {
         mFileProvider = provider;
         break;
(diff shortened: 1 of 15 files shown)


cdm/core/src/main/java/thredds/inventory/MFiles.java
```diff
@@ -6,14 +6,13 @@
...
-import java.util.ServiceLoader;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import thredds.filesystem.MFileOS;
 import thredds.filesystem.MFileOS7;
 import ucar.nc2.internal.ncml.NcmlReader;
+import ucar.nc2.util.NcServiceLoader;
+import ucar.nc2.util.log.LoggerFactory;
```
lesserwhirls: It looks like switching out `org.slf4j.LoggerFactory` for `ucar.nc2.util.log.LoggerFactory` [broke compilation](https://github.com/Unidata/netcdf-java/actions/runs/25727962748/job/75618181913?pr=1555#step:4:144).

lesserwhirls:
I think this approach will work nicely. Thank you so much for your contribution, @nicobrevin! Besides the compilation issue (see above) and the CLA, the only other thing that's tripping up the GitHub tests is the code style. To check the code style, you can run `./gradlew spotlessCheck`, and to apply frixes you can run `./gradlew spotlessApply`. Once the github tests pass, I'll run our full set of tests on our internal Jenkins just to be safe. At that point, we should be good to go and this would get in for the `5.10.0` release.

cdm/core/src/main/java/thredds/inventory/MFiles.java
nicobrevin: Whoops, fixed.  Also ran spotless, but all that did was remove some whitespace from my javadoc :shrug: 

If you think the change needs it, I can add tests, but I gave up quickly when I got errors about missing JDKs.

lesserwhirls: I would not worry about tests for this change, as our existing tests should cover these codepaths quite well. Spotless can be picky about things such as whitespace, which gets annoying. Thanks again for your contribution!

nicobrevin:
> Thank you so much for your contribution, @nicobrevin!

No worries at all, @lesserwhirls , I appreciate your timely response and pragmatism :)

lesserwhirls:
Just a quick follow up that our internal tests also passed!

lesserwhirls:
Just checking in to see if you've had a chance to review the CLA. I'm hoping to get a release out the door by the end of the week, or early next week at the latest, and would be happy to get this in.

PR #1588, opened 2026-07-27
Fix StringLocker behavior.

zliebowitz:
## Description of Changes

Fix issues that could break mutual exclusions properties desired in StringLocker.

## PR Checklist

- [x] Add labels
       until ready for review

### DIFF
--- a/cdm/core/src/test/java/ucar/nc2/StringLockerTest.java
+++ b/cdm/core/src/test/java/ucar/nc2/StringLockerTest.java
@@ -1,0 +1,37 @@
+package ucar.nc2;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+public class StringLockerTest {
+  @Test
+  public void testNonconflicting() throws InterruptedException {
+    Thread t = new Thread(() -> {
+      StringLocker locker = new StringLocker();
+      locker.control("first");
+      locker.control("second");
+      locker.release("first");
+      locker.release("second");
+    });
+    t.start();
+    t.join(500);
+    Assert.assertFalse(t.isAlive());
+  }
+
+
+  @Test
+  public void testNonconflictingOtherOrder() throws InterruptedException {
+    Thread t = new Thread(() -> {
+      StringLocker locker = new StringLocker();
+      locker.control("first");
+      locker.control("second");
+      locker.release("second");
+      locker.release("first");
+    });
+    t.start();
+    t.join(500);
+    Assert.assertFalse(t.isAlive());
+  }
+}
--- a/cdm/core/src/main/java/ucar/nc2/StringLocker.java
+++ b/cdm/core/src/main/java/ucar/nc2/StringLocker.java
@@ -1,8 +1,7 @@
 package ucar.nc2;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
+import java.util.concurrent.ConcurrentSkipListSet;
 
 /**
  * A list of strings that only allows one thread to use any given value at the same time.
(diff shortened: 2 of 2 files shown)


zliebowitz:
This pull request demonstrates the issue (via test) and fixes via fixing StringLocker.

I'm not sure if you'd rather go after this issue in a different fashion (removing the StringLocker), but I figure this at least shows the bug pretty directly.

cdm/core/src/test/java/ucar/nc2/StringLockerTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Test
+  public void testConflicting() {
+    Thread t = new Thread(() -> {
+      StringLocker locker = new StringLocker();
+      locker.control("first");
+      locker.control("second");
+      locker.release("second");
+      locker.control("first");
+      Assert.fail("first has been locked twice");
```
lesserwhirls: I cannot get this to fail on current main, and I don't know that it will. Since the control/release calls in the new `Thread` are done serially, the current implementation of `StringLocker` will wait until another thread calls release on `first` before hitting the `Assert.fail()` call. Since that does not happen, we will always hit the 50 millisecond max wait on the call to join, at which point the test continues to run and passes while the new Thread waiting on the second call of control for `first` sits active (but waiting). I think we'd need a test with multiple threads calling control and release in parallel. It could be quite difficult to actually trigger this specific case (locking a string twice) reliably. That said, I can 100% see where this could happen, but given the code changes above, would no longer be possible. I'm good with removing this test all together.

cdm/core/src/test/java/ucar/nc2/StringLockerTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runners.model.TestTimedOutException;
+
+public class StringLockerTest {
+  @Test
+  public void testNonconflicting() {
+    StringLocker locker = new StringLocker();
+    locker.control("first");
```
lesserwhirls: It looks like purpose of this test, as well as `testNonconflictingOtherOrder`, is to make sure calls to control/release complete. The failure mode would be if, for whatever reason, the second call to control hangs in a wait state. If that is the case, then in both cases, I would run the sequence of locker calls in a new Thread. Then, as with `testConflicting`, I would call join on the thread with a `50l` timeout. The actual test would be that the thread object's `.isActive()` method would return false after the join. If it returns true, then that means something is hanging onto the thread keeping it alive (and `50 ms` should be plenty of time to make those control/release calls and complete the thread activity).

lesserwhirls (changes requested):
Thank you for your contribution! Overall the changes look great! Just a few comments regarding the tests.

cdm/core/src/test/java/ucar/nc2/StringLockerTest.java
zliebowitz: Yeah. Sorry on that, I thought I had tested something close enough on main, but I pushed a test that does seem to capture the bug locally.

If this appears flakey or you have any nervousness surrounding it, I've no issue removing the test.

cdm/core/src/test/java/ucar/nc2/StringLockerTest.java
zliebowitz: Yep. I just wanted to get some minimal test coverage to make sure I didn't completely break things (having the other test exposing the bug verify that we do hit locking cases).

cdm/core/src/test/java/ucar/nc2/StringLockerTest.java
lesserwhirls: Let's just remove this one test now and then we can get your PR merged.

lesserwhirls (approved):
Thanks so much @zliebowitz!

PR #1574, opened 2026-06-23
Add Grib 41

SebastianThomas:
## Description of Changes

This PR only adds support for the 4.41 Template type. It is basically a copy of the other types, but following the structure doc 4.41: https://www.cosmo-model.org/content/model/cosmo/grib/pdtemplate_4.41.htm

Has been tested on a File provided by MeteoSwiss: [ICON CH2 OGD](https://data.geo.admin.ch/browser/index.html#/collections/ch.meteoschweiz.ogd-forecasting-icon-ch2) -> Parameter POACsnc (CTRL), e.g., [here](https://data.geo.admin.ch/browser/index.html#/collections/ch.meteoschweiz.ogd-forecasting-icon-ch2/items/06222026-1200-0-poacsnc-ctrl-qxn4qn26?.asset=asset-icon-ch2-eps-202606221200-0-poacsnc-ctrl-grib2) (may be out of date when reviewing). 

Disclosure: The test case has been generated with AI following the structure of the PDS 4.40 test case. 

## PR Checklist

- [ ] Link to any issues that the PR addresses
- [x] Add labels
- [x] Open as a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/)
       until ready for review
- [x] Make sure GitHub tests pass
- [x] Mark PR as "Ready for Review"

PR #1590, opened 2026-08-10
Add support for template 3.33

allanfrankwork:
## Description of Changes

Add support for grib files using grid definition template 3.33 (https://library.wmo.int/viewer/35625/download?file=WMO-306-v-I-2-2025_en.pdf&type=pdf&navigator=1 page 40.

GDS template 3.33 is the same as 3.30 except 4 additional values (nux, ncx, nuy and ncy).

I know of no public grib files that use 3.33 yet, so testing was done internally - and as I understand it, these 4 values are only used to be able to place the grib file in a larger context (so not used .

But even if those values are not used, parsing still fails with "Unsupported GDS type = 33" - therefore this PR.

## PR Checklist

- [x] Add labels
       until ready for review

### DIFF
--- a/grib/src/main/java/ucar/nc2/grib/grib2/Grib2Gds.java
+++ b/grib/src/main/java/ucar/nc2/grib/grib2/Grib2Gds.java
@@ -52,6 +52,9 @@ public abstract class Grib2Gds {
       case 30:
         result = new LambertConformal(data, 30);
         break;
+      case 33:
+        result = new LambertConformalWithModellingSubdomains(data, 33);
+        break;
       case 31:
         result = new AlbersEqualArea(data);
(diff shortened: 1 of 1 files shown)


allanfrankwork:
Labels grib2 and iosp: grib could be added.

lesserwhirls:
Thank you for your contribution! If you would not mind, could you let us know when a sample message would be available?

allanfrankwork:
Yes, I expect end of this year.

I had hoped to do something similar to https://github.com/Unidata/netcdf-java/pull/1574 for this once the index can be generated.

lesserwhirls:
Excellent, thank you @allanfrankwork!

PR #1591, opened 2026-08-11
Expose `Grib2JpegDecoder`

archmoj:
## Description of Changes

- Change `Grib2JpegDecoder` constructor to public access modifier

Exposing `Grib2JpegDecoder` constructor could help applications to extract data section in JPEG-2000 format directly.
Also according to this documentation (https://docs.unidata.ucar.edu/netcdf-java/4.6/javadocAll/index.html?ucar/nc2/grib/grib2/Grib2JpegDecoder.html) it is expected to be a public method.
Please let me know if I'm missing something.
Thank you.

## PR Checklist

- [ ] Add labels
       until ready for review

### DIFF
--- a/grib/src/main/java/ucar/nc2/grib/grib2/Grib2JpegDecoder.java
+++ b/grib/src/main/java/ucar/nc2/grib/grib2/Grib2JpegDecoder.java
@@ -69,7 +69,7 @@ public class Grib2JpegDecoder {
    * @throws IllegalArgumentException If 'argv' is empty
    * @see Grib2JpegDecoder#getExitCode
    */
-  Grib2JpegDecoder(int nbits, boolean debug) {
+  public Grib2JpegDecoder(int nbits, boolean debug) {
     this.rate = nbits;
     this.debug = debug;
 


lesserwhirls:
Greetings @archmoj! Thank you for your suggestion. I am open to the idea of opening up access to this class, with the understanding that it isn't part of the actual [netCDF-Java public API](https://docs.unidata.ucar.edu/netcdf-java/4.6/javadoc/index.html?overview-summary.html), and it isn't subject to the same guarantees implied with semantic versioning (the `javadocAll` javadocs are really meant for netCDF-Java developers, not application developers). That said, if a developer needs to call this, I'm guessing it is likely because a specific GRIB Grid Definition Section isn't supported (and thus netCDF-Java bails early), and I would _much rather_ get that supported so that data can actually be read through the public API.

Can you tell me a little more about the specific use case you have for wanting to use this directly?

archmoj:
Hi @lesserwhirls,
Thanks for the review and all your great contributions!
I'm working on this App where I handled various GRIB2 sections. But for the data section when it is encoded in JPEG-2000 format I need to use an external library.
For the moment I imported `ucar.jpeg.jj2000.j2k.decoder.Grib2JpegDecoder` and used its `Grib2JpegDecoder` here:
https://github.com/archmoj/grib2_solarchvision/blob/29325e31a2cd7ac20d15232de9526435544c6294/app/src/main/java/grib2_solarchvision/App.java#L9501-L9507 

```java
          String[] argv = new String[4];
          argv[0] = "-rate";
          argv[1] = nf(this.NumberOfBitsUsedForEachPackedValue, 0); // number of bits per pixel
          argv[2] = "-verbose";
          argv[3] = "off";

          Grib2JpegDecoder g2j = new Grib2JpegDecoder(argv);
```

But it would be nice to replace it with this adaptation of `jj2000.j2k.decoder.Decoder`, in order to read input from memory.

Also it would simplify providing the dependencies here:
https://github.com/archmoj/grib2_solarchvision/blob/29325e31a2cd7ac20d15232de9526435544c6294/app/build.gradle#L32-L34
by removing the need to provide `ibs/grib-4.3.jar`.

Thank you!

lesserwhirls:
Having looked at bit closer at your repo, I think there is a much better path forward for you. The grib jar you have is from a 2011 release of netCDF-java (the current release version of both cdm-core and grib is 5.10.0). I agree getting rid of the old grib jar is a worthy goal. Since the time of 4.3 of the grib jar, the jj2000 code has been broken out into its own repository and its own artifact. I would recommend two step to move forward:

1. Remove the dependency on your local copy of `grib-4.3.jar` from your build and replace it with `implementation 'edu.ucar:jj2000:5.4'` (you should also consider upgrading `cdm-core` to version 5.10.0).
2. Copy the `Grib2JpegDecoder` class from https://github.com/Unidata/netcdf-java/blob/maint-5.x/grib/src/main/java/ucar/nc2/grib/grib2/Grib2JpegDecoder.java and use that without fear of relying on a non public (or public, but only for internal use) API. Since the `Grib2JpegDecoder` is licensed under BSD-3, there should not be any issues as long as you follow the very basic terms outlined [in the license](https://github.com/Unidata/netcdf-java/blob/maint-5.x/LICENSE).

Cheers!

PR #1602, opened 2026-09-15
grib2: do not emit Grib2_Statistical_Process_Type attribute when Variable has no intervals

### LINKED ISSUE #1600: grib2: "Grib2_Statistical_Process_Type" attribute has "UnknownStatType--1" for non-interval products

5.10.0
5.11.0 SNAPSHOT

### What went wrong?

A GRIB2 Variable with no time intervals always has a `Grib2_Statistical_Process_Type` attribute with the value `UnknownStatType--1`. (This happens because the string format is "UnknownStatType-%d" and the substitution is `-1`).

Although a minor issue, it's annoying and can cause confusion. I first thought this was out of date lookup tables, then concluded `UnknownStatType--1` is (unintentionally) a special string which means "no intervals".

The description of unrelated issue #730 shows exactly this happening.

### Relevant log messages

_No response_

No

### Code of Conduct

- [x] I agree to follow the UCAR/Unidata Code of Conduct

cambecc:
## Description of Changes

Fixes #1600.

The gist of this change is to:
  - guard emitting of this attribute with `getIntvType() >= 0`
  - annotate `getStatisticName()` as `@Nonnull` and use that to simplify the places it's called

## PR Checklist

- [ ] Add labels
       until ready for review


grib/src/main/java/ucar/nc2/grib/collection/Grib2Iosp.java
```diff
@@ -109,20 +107,14 @@ static String makeVariableLongName(Grib2Tables cust, GribCollectionImmutable.Var
...
+          f.format(" (%s %s)", intvName, statName);
         } else {
-          f.format(" (%s %s)", vindex.getIntvName(), intvName);
+          f.format(" (%s)", statName);
         }
-
-      } else if (vindex.getIntvType() >= 0) {
-        String intvName = cust.getStatisticNameShort(vindex.getIntvType());
-        f.format(" (%s)", intvName);
       }
```
cambecc, line 118: this block of logic becomes simpler now that `getStatisticNameShort` is `@Nonnull`

grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Tables.java
```diff
@@ -244,9 +247,10 @@ public String getStatisticName(int id) {
     return result;
   }
 
+  @Nonnull
   public String getStatisticNameShort(int id) {
     GribStatType stat = GribStatType.getStatTypeFromGrib2(id);
-    return (stat == null) ? "UnknownStatType-" + id : stat.toString();
+    return (stat == null) ? String.format("UnknownStatType%d", id) : stat.toString();
   }
```
cambecc, line 254: Slight change here to use the same pattern as "UnknownAerosolType%d"

grib/src/main/java/ucar/nc2/grib/grib2/table/NcepLocalTables.java
```diff
@@ -381,18 +384,17 @@ public GribStatType getStatType(int id) {
...
     if (id < 192)
       return super.getStatisticName(id);
     if (statName == null)
       statName = initTable410();
-    if (statName == null)
-      return null;
-    return statName.get(id);
+    String result = statName.get(id);
+    return result != null ? result : getStatisticNameShort(id);
   }
```
cambecc, line 395: Previously, `getStatisticName` would return `null` if either the table failed to load or the table did not have a value. (It would also attempt to load the table every time it was called.) This change falls back to `getStatisticNameShort`, which falls back to emitting "UnknownStatType" if the short stat name is not one of the hard-coded values, which I think is okay behavior.

lesserwhirls: Sounds good to me.

lesserwhirls (approved):
Thank you for your contribution!

